PR #342, opened 2022-02-08
Delete netlify.toml

ryan-henness-trimble:
Hey Christian, was curious whether we need this file anymore since we are building/deploying from the GitHub side now? Not sure if there is another way to set up that lighthouse config though


coliff:
The main thing Netlify is useful for is the auto-deploy previews of PRs. We have that turned off at the moment but we should find a way to turn it back on. I think we should keep the netlify.toml for now because it's used for branch previews.

coliff:
I think that file is still being used now anyway - without it we wouldn't get this A+ rating.
https://securityheaders.com/?q=modus-web-components.trimble.com&hide=on&followRedirects=on

PR #387, opened 2022-03-07
Use user initials if avatar image is broken

ryan-henness-trimble:
Use user initials if avatar image is broken


yuenm18 (approved):

yuenm18:
you might want to run npm install to so that the version in the package lock file gets updated too

yuenm18 (approved):

Issue #166, opened 2021-12-02
[1] Add Data Table component (barebones)


ryan-henness-trimble:
https://docs.google.com/document/d/1JwHy6Hqrj5EJY6deM5382STvM4InvV3OTdci4f1b9Ac/edit#

We may want to wrap a third party table library. Implementation details are TBD

PR #643, opened 2022-08-15
Barebones Modus Data Table

ryan-henness-trimble:
## Description

Table reference: https://modus.trimble.com/components/tables/

This PR includes the implementation of the barebones Data Table Stencil component, tests, and additional Storybook docs/bug fixes. 

The Data Table is currently a basic implementation that takes columns, data, size, and sort options. 
<details style="color: blue;">
  <summary>🔵 Click for Details, Inputs, Sorting</summary>

### Implementation Details
- The table `columns` property accepts an array of strings, or an array of objects.
  - Defining the columns with objects allows you to define extra properties on the column/column cells.
- The table `data` property accepts either a 2D array of primitive types (`string`, `number`, or `boolean`) or an array of objects.
  - Note: the objects in `data` get their property names mapped to each column `id`.
- The table needs a unique `id` for each column that maps to each cell in that column. If it is not provided, the table will generate a unique id for each column.

### Inputs
Refer to the **Default** example above to see how to define simple `columns` and `data` in the table.

To define columns and data in the table *using objects*:
```html
<modus-data-table />
<script>
    document.querySelector('modus-data-table').columns = [
        { display: 'Name', id: 'name', width: '50%' },
        { display: 'Age', id: 'age', align: 'right' },
        { display: 'Contacted', id: 'contacted' }
    ];

    const john = { name: 'John', age: 25, contacted: false };
    const jane = { name: 'Jane', age: 26, contacted: false };
    const joe = { name: 'Joe', age: 27, contacted: true };
    document.querySelector('modus-data-table').data = [john, jane, joe];
</script>
```

### Sorting
- The table uses a tri-state for its sorting:
  - `none`: no sorting is applied
  - `asc`: ascending sort
  - `desc`: descending sort
- The table sorting options can be set with the optional `sortOptions` property. The property accepts a `ModusTableSortOptions` object.
  - The object defines whether the table can be sorted on header click, and if the sort is done on the client or server-side.
- If the table is sortable, it will output the `sort` event on header click. This event's detail is a `ModusDataTableSortEvent` object.
- If the table's `serverSide` sort option is `false`, it will sort the columns on the client-side. Otherwise, it is up to the consumer to update the table's `data` property with the sorted data.

</details>

Inline stories were broken for Storybook stories that needed properties set through a script tag, I set those stories' `inlineStories` to false.

References #166 

## Type of change

- [X] Documentation update

## How Has This Been Tested?

Manually, unit, and e2e.

## Checklist:

- [X] I have made corresponding changes to the documentation (including CHANGELOG updates)
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ View all releases at: https://github.com/trimble-oss/modus-web-components/releas
 
 ## Unreleased
 
+### Added
+- Barebones Modus Data Table
+
 ### Fixed
 
 - Modus Chips active state color is now blue
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -98,6 +98,8 @@ For more information about Stencil's testing, and a better distinction between t
 
 Run the tests with `npm run test`.
 
+To run a specific test, run `stencil test --spec --e2e --silent [test file name]`.
+
 ## Making Changes and Submitting a PR
 
 1. Before working on an issue, the repository should be forked with intent to contribute to the parent repository.
--- a/README.md
+++ b/README.md
@@ -9,9 +9,6 @@
   <a href="https://www.npmjs.com/package/@trimble-oss/modus-web-components">
     <img src="https://img.shields.io/github/package-json/v/trimble-oss/modus-web-components?color=blue&filename=stencil-workspace%2Fpackage.json" alt/>
   </a>
-  <a href="https://app.netlify.com/sites/modus-web-components/deploys">
-    <img src="https://api.netlify.com/api/v1/badges/c9f1de7d-daf8-4dd4-876d-4aa36a077213/deploy-status" alt/>
-  </a>
   <a href="https://github.com/trimble-oss/modus-web-components/graphs/contributors">
     <img src="https://img.shields.io/github/contributors/trimble-oss/modus-web-components?color=lightblue" alt/>
   </a>
--- a/angular-workspace/projects/trimble-oss/modus-angular-components/src/lib/stencil-generated/components.ts
+++ b/angular-workspace/projects/trimble-oss/modus-angular-components/src/lib/stencil-generated/components.ts
@@ -246,6 +246,34 @@ export class ModusChip {
   }
 }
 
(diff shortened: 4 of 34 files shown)


ryan-henness-trimble:
@coliff I had a question about the 'readonly' cells. If the table is completely read only, should we still display the cells in a read only style? It seems like it might be better to display them as 'normal' cells with a white background if its completely read only

msankaran0712:
> @coliff @msankaran0712 I had a question about the 'read-only' cells. If the table is completely read-only, should we still display the cells in a read-only style? It seems like it might be better to display them as 'normal' cells with a white background if its completely read-only. It's looking odd to me with only 'read-only' cells in the table so I wanted to double check
> 
> Read-only styling image
> 
> Normal styling image

A read-only table is just a normal table, the cells should not be disabled. You can refer the List table styles in https://modus.trimble.com/components/tables/?q=tabl#types

msankaran0712:
Also are we doing storybook changes in a separate PR ?

ryan-henness-trimble:
> Do we have a preview for these changes?



> > @coliff @msankaran0712 I had a question about the 'read-only' cells. If the table is completely read-only, should we still display the cells in a read-only style? It seems like it might be better to display them as 'normal' cells with a white background if its completely read-only. It's looking odd to me with only 'read-only' cells in the table so I wanted to double check
> > Read-only styling image
> > Normal styling image
> 
> A read-only table is just a normal table, the cells should not be disabled. You can refer the List table styles in https://modus.trimble.com/components/tables/?q=tabl#types

There's two different styles for read-only and normal in the docs

ryan-henness-trimble:
Opening branch PR for static web app deployment

msankaran0712:
@ryan-henness-trimble I think this PR should be moved to `Draft` if you have more commits.

ryan-henness-trimble:
> @ryan-henness-trimble I think this PR should be moved to `Draft` if you have more commits.

it's ready 😄 I just had some small changes

quynh-le-trimble (approved):
Nice job on all the tests.

PR #724, opened 2022-09-12
Add `package.json` to root

coliff:
This file is needed so we can download the entire repo to the style guide site so we can re-use the markdown docs.
This doesn't needed to ne mentioned in CHANGELOG.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "modus-web-components",
+  "version": "1.0.0",
+  "private": true,
+  "description": "Modus Web Components",
+  "homepage": "https://modus-web-components.trimble.com/",
+  "bugs": {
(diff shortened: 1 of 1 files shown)


coliff:
Marking as draft for now. I don't need this right now as I found a workaround using a 3rd party solution by adding this to my `package.json`:

`"modus-web-components": "https://gitpkg.now.sh/trimble-oss/modus-web-components/stencil-workspace?main",`

ryan-henness-trimble:
hmm, do you think we still need this or is the 3rd part solution alright?

coliff:
> hmm, do you think we still need this or is the 3rd party solution alright?

The 3rd party solution is alright for now... I'll close this PR - we can always re-open in the future if that https://gitpkg.vercel.app/ service goes offline.

PR #814, opened 2022-10-24
Issue 808/modus tabs component could be dynamic resizable

### LINKED ISSUE #808: modus-tabs component could be dynamic resizable.
**Describe the solution you'd like**
I would like to use this component in a browser panel. This panel should be resizable by user and therefore any other component from inside this panel should be resizable as well. 
Right now modus-tab component is not resizable, it has a specific length which cannot be altered by user or any other components.

DragosTataru:
## Description

The "width" property for modus-tabs was added in order to determine the component to change it's width when the parent width is changing. For example when the parent is resized (could be a panel in which the modus-tabs is inserted). 
For this change there are no extra required dependencies.

Fixes #808

## Type of change

- [x] Documentation update

## How Has This Been Tested?

It was tested by using component modus-tabs in `index.html` alongside with some `<div>` parents
also the following script was used: (see attached txt file)

## Checklist:

- [x] - [ ] I have checked my code and corrected any misspellings

[Testing in Modus modus-tab component.txt](https://github.com/trimble-oss/modus-web-components/files/9851298/Testing.in.Modus.modus-tab.component.txt)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,13 @@ All notable changes to this project will be documented in this file.
 
 View all releases at: https://github.com/trimble-oss/modus-web-components/releases/
 
-### Added
-
-- Added "width" for modus-tabs in CSS file. Now component can be resizable from UI.
-
 ## Unreleased
+
+### Added 
+
+- Added optional property `fullWidth` for modus-tabs. Now component can be resizable from UI if `fullWidth` flag is set on true.
+
+## 0.1.23 - 2022-10-26
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ View all releases at: https://github.com/trimble-oss/modus-web-components/releas
 
 ## Unreleased
 
+## 0.1.22 - 2022-10-18
+
+### Fixed
+
+- Fixed an issue with the Data Table where setting column ids was not working
+
 ## 0.1.21 - 2022-10-04
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ View all releases at: https://github.com/trimble-oss/modus-web-components/releas
 
 ## Unreleased
 
+### Added
+
+- Added a theme toggler to the toolbar in storybook pages to switch between `light` and `dark` themes
+
 ## 0.1.22 - 2022-10-18
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 View all releases at: https://github.com/trimble-oss/modus-web-components/releases/
 
+### Added
+
+- Added "width" for modus-tabs in CSS file. Now component can be resizable from UI.
+
 ## Unreleased
 
 ### Added
--- a/stencil-workspace/src/components/modus-tabs/modus-tabs.spec.ts
+++ b/stencil-workspace/src/components/modus-tabs/modus-tabs.spec.ts
@@ -24,4 +24,10 @@ describe('modus-button', () => {
     className = modusTabs.classBySize.get('small');
     expect(className).toEqual('small');
   });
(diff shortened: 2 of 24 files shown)


coliff:
> @coliff What are your thoughts on a `fullWidth` flag input property? If `true`, the `width: 100%` is set on the tabs. Otherwise, the functionality remains the same.
> 
> I think this is how I would handle this issue. That would also make sure that it does not break any existing usages

Really, we should stick to the Modus specs and not offer an option to break the rules. Adding `width:100%` to modus-tabs would not be a breaking change so we should definitely keep that. I would vote to not have the option for `width:100%` on individual tabs, but if you feel strongly about allowing that then we should have that as opt-in and not default.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 View all releases at: https://github.com/trimble-oss/modus-web-components/releases/
 
+### Added
```
ryan-henness-trimble, line 7: This should be moved down un the `## Unreleased` section

DragosTataru:
> 

Thank you for the feedback @coliff . I ran some tests without adding width:100% to the tabs themselves but now resizing is not longer possible. Indeed, adding width value to tabs it's against modus specs for this component but I'm wondering if there is no other way to make it resizable. I'm looking for a particular behavior for modus-tabs component, something similar to this GIF 
resize for panel

coliff:
@DragosTataru heya- I'm just checking on this and will get back to you later today. The design in that GIF looks fine and I've seen that design in many other web apps/designs, but it isn't currently something we have in Modus. I'll check if we can add it as an opt-in option. :-)

coliff:
I heard back, we can include the width:100% on the tabs, but it should be opt-in and not default... so I think we should do what Ryan suggested and have that `fullWidth` flag input property. If true, the width: 100% is set on the tabs, if not set or false it'll default to standard fixed-width behavior.

DragosTataru:
> I heard back, we can include the width:100% on the tabs, but it should be opt-in and not default... so I think we should do what Ryan suggested and have that `fullWidth` flag input property. If true, the width: 100% is set on the tabs, if not set or false it'll default to standard fixed-width behavior.

Sounds good to me, I will add the flag and update the PR

stencil-workspace/storybook/stories/components/modus-tabs/modus-tabs-storybook-docs.mdx
```diff
@@ -57,3 +58,4 @@ A new TypeScript typing has been provided named Tab defined as:
 
 - Tabs gets an `aria-label` provided by the `aria-label` property input.
 - When a Tab has focus, **Enter** activates the Tab, and emits the `tabChange` event.
+- Set `full-width` on `true` for Tab component in order cu make it resizable inside a parent.
```
coliff, line 61: typo here- 'in order **cu** make it'

coliff (approved):

coliff:
We need to add a demo/example of this to the docs really, but can do that in a separate PR.

stencil-workspace/src/components/modus-tabs/modus-tabs.spec.ts
```diff
@@ -24,4 +24,10 @@ describe('modus-button', () => {
     className = modusTabs.classBySize.get('small');
     expect(className).toEqual('small');
   });
-});
+  it('should have flag "fullWidth" set on false as default', async () => {
```
ryan-henness-trimble, line 27: Let's add a new line before in between this test and the one above it

stencil-workspace/src/components/modus-tabs/modus-tabs.tsx
```diff
@@ -13,6 +13,8 @@ export interface Tab {
   shadow: true,
 })
 export class ModusTabs {
+  /*(false as default) the tab full width, can be resized inside a parent*/
```
ryan-henness-trimble, line 16: Let's update the comment on this one, and make sure there is a new line in between this and the next `@Prop`

```
/* Whether the Tabs take up the full width of their parent container. */
@Prop() fullWidth = false;
```

PR #837, opened 2022-11-02
Custom styling POC

ryan-henness-trimble:
Hey everyone - this is a POC for custom element styling. The idea here is to inject a custom style string into the shadow DOM's style tag. This idea may need some more solidification/checking, but it's at least good to know there may be an option here

Here's a look at the component in the dev env:
image

### DIFF
--- a/stencil-workspace/src/index.html
+++ b/stencil-workspace/src/index.html
@@ -21,6 +21,15 @@
   </head>
   <body>
       <!-- Test your components here -->
-      <modus-badge color="secondary" custom-style="div.badge { background-color: red !important; border-radius: 16px; padding: 10px; }">Secondary</modus-badge>
+      <modus-badge color="secondary">
+        <style slot="custom-style-with-slot">
+          div.badge {
+            background-color: red !important;
+            border-radius: 16px;
+            padding: 10px;
+          }
+        </style>
+        1234
+      </modus-badge>
   </body>
 </html>
--- a/stencil-workspace/src/index.html
+++ b/stencil-workspace/src/index.html
@@ -21,5 +21,6 @@
   </head>
   <body>
       <!-- Test your components here -->
+      <modus-badge color="secondary" custom-style="div.badge { background-color: red !important; border-radius: 16px; padding: 10px; }">Secondary</modus-badge>
   </body>
 </html>
--- a/stencil-workspace/src/components.d.ts
+++ b/stencil-workspace/src/components.d.ts
@@ -66,6 +66,7 @@ export namespace Components {
           * (optional) The color of the badge
          */
         "color": 'danger' | 'dark' | 'primary' | 'secondary' | 'success' | 'tertiary' | 'warning';
+        "customStyle": string;
         /**
           * (optional) The size of the badge
          */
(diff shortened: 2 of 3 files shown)


msankaran0712:
@ryan-henness-trimble I understand the ultimate need for custom styling, it can be frustrating for the users when they want to style a component and are unable to do it. 

This looks like a way to achieve custom styling yet I feel it is missing something. 
Maybe we need a better way of injecting the styles rather than passing it as a string, in the HTML and should be able to use SASS. 

Did you get a chance to look at other libraries in the market? Like shoelace, ionic framework,etc.

cc @coliff

coliff:
Nice. I think this is a great way of offering an 'emergency quick fix' for adding custom styling for a particular use case. (like perhaps reducing the horizontal padding of a button by a few pixels so it fits within a div for example).

**My thoughts on offering custom styling:** I don't see the need for offering proper custom styling of components at all, the point of using Modus is that the components should be following the Modus style guidelines and not using custom colors, fonts etc. Shoelace, Bootstrap and Ionic frameworks are designed to be styled by the developers, but Modus Web Components should be ready to use out-of-the-box and custom styling shouldn't really be promoted. If a developer/team wants something added/changed then I think they should request it be added to the Modus Style Guide.

ryan-henness-trimble:
> Nice. I think this is a great way of offering an 'emergency quick fix' for adding custom styling for a particular use case. (like perhaps reducing the horizontal padding of a button by a few pixels so it fits within a div for example).
> 
> **My thoughts on offering more comprehensive custom styling:** I don't see the need for offering proper custom styling of components at all, the point of using Modus is that the components should be following the Modus style guidelines and not using custom colors, fonts etc. Shoelace, Bootstrap and Ionic frameworks are designed to be styled by the developers, but Modus Web Components should be ready to use out-of-the-box and custom styling shouldn't really be promoted. If a developer/team wants something added/changed then I think they should request it be added to the Modus Style Guide.

Yep I agree the library should be ready to use out-of-the-box and that custom styling should be discouraged. But I know from practical use of these components in prod that there are times that we would like to slightly tweak the components that probably shouldn't be considered a style guide change. I think the key here is for us to find a good way to balance/implement what should be considered an emergency change and a style guide change 

I am a bit worried about a couple of things: 
- Trying to reimplement too many common styling properties through Stencil props that devs are used to having access to
- Making it too involved of a process for devs to get these custom changes into the style guide, it may shoo them away from using the library at all

stencil-workspace/src/index.html
```diff
@@ -21,6 +21,15 @@
   </head>
   <body>
       <!-- Test your components here -->
-      <modus-badge color="secondary" custom-style="div.badge { background-color: red !important; border-radius: 16px; padding: 10px; }">Secondary</modus-badge>
+      <modus-badge color="secondary">
+        <style slot="custom-style-with-slot">
```
msankaran0712, line 25: This looks Nice !! better than passing string.

msankaran0712: And we can keep the slot name smaller, maybe just 'custom-style'

PR #848, opened 2022-11-09
Dark mode for Dropdown, List and File Upload Drop Zone

msankaran0712:
Fixes #802 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (including CHANGELOG updates)
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings

PR #850, opened 2022-11-10
Dark mode - Rearrange CSS styles alphabetically

msankaran0712:
Fixes a comment in [PR](https://github.com/trimble-oss/modus-web-components/pull/848#issuecomment-1308651188)


coliff (approved):

coliff:
Nice. I tested a good Stylelint plugin which fixes this automatically.. (when you run `stylelint --fix` I'll take another look and maybe make a PR with that added so we don't need to manually sort them.

PR #914, opened 2022-12-09
Add a test project for React

### LINKED ISSUE #843: Dark mode - verify React output
If possible set up something deployable to view it on a web page.

msankaran0712:
- Created a project inside /react-workspace/test-app 
- Used `create-react-app` for building the project skeleton
- Includes a hacky solution for [importing components from outside test-app/src/ folder issue](https://github.com/facebook/create-react-app/issues/834) for time's sake 
  - The solution was to create an npm pack and move it inside test-app/src everytime when running the command `npm run test-app: start` 

Fixes #843


stencil-workspace/SUPPORT.md
```diff
@@ -0,0 +1,22 @@
+# Support
```
ryan-henness-trimble, line 1: Do we want to include this file in the npm package?

react-workspace/package.json
```diff
@@ -33,22 +36,24 @@
...
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
     "typescript": "^4.7.4"
   },
   "peerDependencies": {
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2"
+    "react": "^18.2.0",
```
ryan-henness-trimble, line 56: We're currently building the react package with React @ `17.0.2`. Makes me wonder if we will need to build out separate test apps for the major versions, like `test-app-17`, and `test-app-18`

ryan-henness-trimble:
We should make sure this is providing enough value for the work it will take, especially if we have to support multiple test apps for different React versions. Now that I'm thinking about it more, it kind of feels like we'd be testing Stencil's framework output implementation rather than the Modus Components. Any thoughts?

PR #1032, opened 2023-02-20
1020 - Update the code to add the aria-label property to the modus-modal's both buttons

### LINKED ISSUE #1020: [BUG] Modal Buttons missing accessibility attributes
**Describe the bug**
Modal Button (primary-button and secondary-button) are missing accessibility attributes

**To Reproduce**

Go to https://modus-web-components.trimble.com/?path=/docs/components-modal--default
Click on 'open modal'
The Close button should have attributes role="button" and aria-label="Close"

1. Go to https://modus-web-components.trimble.com/?path=/docs/components-modal--default
2. Click on 'open modal'
3. Scroll down to the popup model buttons (primary-button ("Sweet!") and secondary-button ("Save changes"))
4. Both buttons are missing aria-label

**Expected behavior**
The modus-modal's both buttons should have the aria-label property for the accessibility

**Screenshots**
image

hrgondaliya:
Update the code to add the aria-label property to the modus-modal's both buttons

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes #1020

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

## Checklist

- [ ] I have made corresponding changes to the documentation (including CHANGELOG updates)


msankaran0712 (changes requested):
We also need to change the properties for the Modal component on the storybook page.

stencil-workspace/src/components.d.ts
```diff
@@ -423,10 +423,18 @@ export namespace Components {
...
+         */
+        "primaryButtonAreaLabel": string | null;
         /**
           * (optional) The modal's primary button text.
          */
         "primaryButtonText": string;
+        /**
+          * (optional) The modal's secondary button Area Label.
+         */
+        "secondaryButtonAreaLabel": string | null;
```
msankaran0712, line 437: I believe it should read  `..AriaLabel`

msankaran0712 (changes requested):

msankaran0712:
It would be nice to have something like https://shoelace.style/components/dialog?id=slots for the footer section as optional, what do you think? @ryan-henness-trimble

stencil-workspace/src/components/modus-modal/modus-modal.tsx
```diff
@@ -23,6 +23,12 @@ export class ModusModal {
   /** (optional) The modal's z-index. */
   @Prop() zIndex = '1';
 
+  /** (optional) The modal's primary button Area label. */
+  @Prop() primaryButtonAreaLabel: string | null;
```
msankaran0712, line 27: I believe it should read ..AriaLabel, also in the description text.

msankaran0712 (changes requested):

stencil-workspace/src/components/modus-modal/modus-modal.tsx
```diff
@@ -23,6 +23,12 @@ export class ModusModal {
...
+
+  /** (optional) The modal's secondary button Area Label. */
+  @Prop() secondaryButtonAreaLabel: string | null;
```
msankaran0712, line 30: I believe it is sufficient to use only `string` for prop type @ryan-henness-trimble please correct me if am wrong.

msankaran0712 (changes requested):

stencil-workspace/src/components/modus-modal/modus-modal.tsx
msankaran0712: I recommend keeping only `string`.

ryan-henness-trimble: If I remember correctly, the Storybook accessibility checks were not passing unless we included `null`

stencil-workspace/src/components/modus-modal/modus-modal.tsx
```diff
@@ -23,6 +23,12 @@ export class ModusModal {
   /** (optional) The modal's z-index. */
   @Prop() zIndex = '1';
 
+  /** (optional) The modal's primary button Aria label. */
```
ryan-henness-trimble, line 26: Can we change `Aria label` to `aria-label` in the comments?

stencil-workspace/src/components/modus-modal/modus-modal.tsx
```diff
@@ -23,6 +23,12 @@ export class ModusModal {
...
+  @Prop() primaryButtonAriaLabel: string | null;
+
+  /** (optional) The modal's secondary button Aria Label. */
+  @Prop() secondaryButtonAriaLabel: string | null;
```
ryan-henness-trimble, line 30: Let's alphasort these `@Prop()`s - that way they're next to their corresponding `ButtonText` prop

ryan-henness-trimble:
> It would be nice to have something like https://shoelace.style/components/dialog?id=slots for the footer section as optional, any thoughts? @ryan-henness-trimble

It depends on the Modus spec, do we want to allow the user to add more than 3 buttons? Do we want them to be able to render elements in the footer other than buttons?

ryan-henness-trimble (approved):

msankaran0712:
> > It would be nice to have something like https://shoelace.style/components/dialog?id=slots for the footer section as optional, any thoughts? @ryan-henness-trimble
> 
> It depends on how rigid we want to adhere to the Modus specs. Do we want to allow the dev to add more than two buttons? Do we want them to be able to render elements other than buttons in the footer?
> 
> If we want to give them full control, a slot sounds like a good option. But if we want to limit what can be rendered in the footer, we might want to consider another option
> 
> I would lean towards adding optional slots that will override the default buttons. More control seems better in this case

@ryan-henness-trimble I agree to have optional slots for buttons. It is true that users should not add buttons more than two, but providing a slot as an option will give users better control over the button configurations instead of adding those configs in the Modal's API, which in the future might become too much.

stencil-workspace/src/components/modus-modal/modus-modal.tsx
msankaran0712: hey, @ryan-henness-trimble @hrgondaliya I tested the storybook page by removing `null` from the properties and keeping only `string` as a type, seems like there is no issue with accessibility, so up to you to decide whether to remove `null` or keep it.

PR #1048, opened 2023-02-23
Added Modus Time Picker component

### LINKED ISSUE #973: Time Picker in Web Components

msankaran0712:
Fixes #973 

## Type of change

- [x] Documentation update

## How Has This Been Tested?
Chrome Browser

## Checklist


coliff:
should this have `maxlength="5"`?

msankaran0712:
> Many systems/countries use 24-hour clock, I don't think we should enforce entering PM or AM.
> 
> I don't really see why using this would ever be preferable over using input `type="time"` which supports 12/24 hour clock, step inputs, localization/internationalization, validation and min and max time. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time

Good question, maybe I went ahead with the assumption that we are not using the native time picker. I will confirm it with @egunther39.

ryan-henness-trimble:
This feels like something a developer could build using the Text Input and Selection components 🤔 I wonder if we should add this formatting/logic as options for the Text Input for the Time portion, and expect devs to use the Selection component for the timezones

msankaran0712:
> This feels like something a developer could build using the Text Input and Selection components 🤔 I wonder if we should add this formatting/logic as options for the Text Input for the Time portion, and expect devs to use the Selection component for the timezones - just my first thought

That's a good idea. We can combine Text input, Number input and Time input. I need a couple more days to do that so I think we can create a separate issue. what do you think @ryan-henness-trimble ?

msankaran0712:
hey, @coliff removed maxLength as it seemed redundant, since we have a specific time format and autoformat options, there is no specific use case to support it, also the native input="time" doesn't have it too. We have manually added maxLength inside the code.

ryan-henness-trimble (approved):

ryan-henness-trimble:
> > This feels like something a developer could build using the Text Input and Selection components 🤔 I wonder if we should add this formatting/logic as options for the Text Input for the Time portion, and expect devs to use the Selection component for the timezones - just my first thought
> 
> That's a good idea. We can combine Text input, Number input and Time input. I need a couple more days to do that so I think we can create a separate issue. what do you think @ryan-henness-trimble ?

Oh hmm - I forgot about the Number input to be honest. I wonder if it would be better to just keep them separate

coliff (approved):

PR #1082, opened 2023-03-02
Issue 1061: Updated modus-modal default backdrop color

### LINKED ISSUE #1061: [BUG] Modus Modal defaults to a completely opaque background which obscures the application
**Describe the bug**
The modus-modal component is displaying a completely opaque background which appears to be caused by a default css variable being set without opacity < 1

**To Reproduce**
Install @trimble-oss/modus-web-components version 0.1.34 and use the modus-alert component
When the modal is opened the background will default to a "black" color.

**Expected behavior**
image

**Screenshots**
image

**Desktop (please complete the following information):**

- OS: Windows 10
- Browser: Chrome

**Additional context**
This appears to have been introduced in:
https://github.com/trimble-oss/modus-web-components/pull/907/files#diff-c7ba294f77a7a677b178a13e5a15115ae171cda436f21fb59ff179e5670badb8

in file modus-modal.scsss

leohaughton:
## Description

This PR updates the default modus-modal backdrop color to be consistent with the default theme color. For existing users of modus-web-components they will now get consistent default UX when upgrading to the latest version.

Prior to this fix, when upgrading to the latest version the default backdrop color for the modus-modal was completely opaque.

Fixes #1061 

## Type of change

- [ ] Documentation update

## How Has This Been Tested?
 - Tested manually by using the component inside the test index.html page with and without the global themes applied in stencil.config.json
 - Added an e2e test to ensure that the styles are being applied correctly from the theme however I was unable to write an e2e test that covers the styles when themes are not applied. There is still a gap there for "default" component behavior.

## Checklist

### DIFF
--- a/stencil-workspace/package.json
+++ b/stencil-workspace/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@trimble-oss/modus-web-components",
-  "version": "0.1.34",
+  "version": "0.1.35",
   "description": "Trimble Modus Web Component Library",
   "homepage": "https://modus-web-components.trimble.com/",
   "bugs": {
--- a/stencil-workspace/src/components/modus-modal/modus-modal.e2e.ts
+++ b/stencil-workspace/src/components/modus-modal/modus-modal.e2e.ts
@@ -129,4 +129,30 @@ describe('modus-modal', () => {
     await page.waitForChanges();
     expect(closed).toHaveReceivedEvent();
   });
+
+  it('has a default overlay background color', async () => {
+    const page = await newE2EPage();
+    await page.setContent(
+      '<modus-modal secondary-button-text="Secondary Text"></modus-modal>'
+    );
+    const modal = await page.find('modus-modal');
+    await modal.callMethod('open');
+
+    // Computed overlay color is returned from page.evaluate()
+
+    const backgroundColor = await page.evaluate(async (): Promise<string> => {
+      const overlay = document
+        .querySelector('modus-modal')
+        .shadowRoot.querySelector('.overlay');
+      overlay.scrollIntoView({ block: 'center', inline: 'center' });
+
+      const styles = window.getComputedStyle(overlay);
+      return styles.backgroundColor;
+    });
+
+    await page.waitForChanges();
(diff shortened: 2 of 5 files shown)


stencil-workspace/src/components/modus-modal/modus-modal.e2e.ts
```diff
@@ -129,4 +129,30 @@ describe('modus-modal', () => {
...
+        .shadowRoot.querySelector('.overlay');
+      overlay.scrollIntoView({ block: 'center', inline: 'center' });
+
+      const styles = window.getComputedStyle(overlay);
+      return styles.backgroundColor;
+    });
+
+    await page.waitForChanges();
+    console.error(backgroundColor);
+    expect(backgroundColor).toBe('rgba(37, 42, 46, 0.75)');
```
ryan-henness-trimble, line 155: Could we check against the same hex value as the .scss here? Or is that not possible since we're evaluating the page and getting the computed style?

stencil-workspace/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@trimble-oss/modus-web-components",
-  "version": "0.1.34",
+  "version": "0.1.35",
```
ryan-henness-trimble, line 3: Can we run `npm i` from the `stencil-workspace` too? This will update the `package-lock.json` for us to include this version bump

stencil-workspace/src/components/modus-modal/modus-modal.e2e.ts
leohaughton: Yeah I was thinking the same thing. This is what was coming back from the computed style and I didn't put much more thought into it beyond that.

I could probably figure out how to either convert the rgba back into a hex, or even better, use the sass vars file as the source of truth but I would question the cost / benefit of doing this.

stencil-workspace/package.json
leohaughton: Yep. My bad missed that step.

leohaughton: Done

ryan-henness-trimble (approved):

PR #1220, opened 2023-04-24
Improve Input Helper Text styles

coliff:
- Change helper text from 10px to 11px
- Change helper text weight from normal to **600**
- Docs: Improve labels for examples


msankaran0712 (dismissed):
@coliff I think we need to get this change done in all the user input types such as date, time, select, number, etc.

coliff:
> @coliff I think we need to get this change done in all the user input types such as date, time, select, number, etc.

Good point! I updated the SCSS files for the other components now too.

PR #1296, opened 2023-05-17
Add Compact Progress Bar variant (4px height)

coliff:
Note: The small variant on the site right now was referred to compact in some places.
There is now both a small (8px) and compact variant (4px). 

PREVIEW: https://icy-ground-0374a1310-1296.centralus.1.azurestaticapps.net/?path=/docs/components-progress-bar--default

### DIFF
--- a/stencil-workspace/src/components.d.ts
+++ b/stencil-workspace/src/components.d.ts
@@ -710,7 +710,7 @@ export namespace Components {
         /**
           * (optional) The progress bar's size.
          */
-        "size": 'default' | 'compact';
+        "size": 'default' | 'small' | 'compact';
         /**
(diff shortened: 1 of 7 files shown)


msankaran0712 (approved):

ryan-henness-trimble:
@msankaran0712
I was just taking a peek to see if we had any `breaking-change` merges so that I could get started on the update guide doc

I don't think this qualifies as a breaking change since it was just an addition, current usages of the component don't have to make any changes to keep using it

coliff:
It's sort of a breaking change... Before this was merged the small variant was called compact in some places and small in others. If a user was using compact before (when it was 8px) and continued using that with this change in place the progress bar would then be 4px.

Issue #557, opened 2022-07-11
Add Storybook Storysource Addon


coliff:
This looks useful: https://storybook.js.org/addons/@storybook/addon-storysource/

coliff:
not got it working quite right....

![image](https://user-images.githubusercontent.com/1212885/178245512-e495cf52-b0c3-428f-98d7-04e40db80719.png)

![image](https://user-images.githubusercontent.com/1212885/178245680-f1f2b35f-0419-4324-afb2-6cb207440931.png)

PR #1321, opened 2023-05-22
Add Storybook Storysource Addon

coliff:
Adds this plugin: https://storybook.js.org/addons/@storybook/addon-storysource

Fixes: #557

I don't know if the code it displays is helpful / correct - and if it isn't correct then I don't know how to fix.


coliff:
This doesn't look right - missing syntax highlighting and output looks different compared to the example screenshots at https://storybook.js.org/addons/@storybook/addon-storysource:
image

coliff:
I'll close this for now and maybe revisit again sometime in the future.

Issue #1345, opened 2023-05-25
Modus Navbar's main menu does not take up full screen height


ryan-henness-trimble:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have validated any HTML to avoid common problems
- [X] I have read the [contributing guidelines](https://github.com/trimble-oss/modus-web-components/blob/main/CONTRIBUTING.md)

### Describe the issue

The navbar's main menu (hamburger) on the left should take up the remaining screen height 

### Which npm package?

@trimble-oss/modus-web-components

### What version of npm package are you using?

v0.1.42

### Priority

High

### What product/project are you using Modus Components for?

Trimble Work Center

### What is your team/division name?

Trimble Viewpoint

PR #1346, opened 2023-05-26
Fix NavBar Main menu full screen height

msankaran0712:
## Description

Reverted the change that removed full screen height on NavBar in a [commit](https://github.com/trimble-oss/modus-web-components/commit/edad227c15f9116344a8da272add8677b1f833c1#diff-3fbcd5eb11fa2bbc82440513ab25928befa8beea91da49aea1cd36e7228fa0de)  

@ryan-henness-trimble  I have set the height to `calc(100vh - 58px);`  to avoid scroll bar, earlier it was `calc(100vh - 54px);` because of which a scroll bar was appearing even after removing unwanted spaces from the web page. 
I assume that 54px represents the fixed height on NavBar container.

### created the PR to revert my changes, please feel free to close it if not required.

References #1345 

## Type of change

- [x] Documentation update

## Checklist

### DIFF
--- a/stencil-workspace/src/components/modus-navbar/main-menu/modus-navbar-main-menu.scss
+++ b/stencil-workspace/src/components/modus-navbar/main-menu/modus-navbar-main-menu.scss
@@ -7,6 +7,7 @@
   border-top: none;
   clip-path: inset(0 -10px -10px -10px);
   cursor: default;
+  height: calc(100vh - 58px);
   left: 0;
   max-width: 616px;
   min-width: 248px;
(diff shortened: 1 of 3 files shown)


ryan-henness-trimble:
We might want to take more time with this issue to rethink how we're rendering the navbar menus

The navbar code is making an assumption that the navbar is always going to be used at the top left of the page with 0 padding/margin. This makes me think the menus are missing some encapsulation, we should be able to place things above the header while still having the main menu span only the rest of its container

msankaran0712:
> We might want to take more time with this issue to rethink how we're rendering the navbar menus
> 
> The navbar code is making an assumption that the navbar is always going to be used at the top left of the page with 0 padding/margin. This makes me think the menus are missing some encapsulation, we should be able to place things above the header while still having the main menu span only the rest of its container 🤔

closing the PR for now then 👍🏼

PR #1381, opened 2023-06-05
Skip Azure Static Web App deployment on missing secrets

ryan-henness-trimble:
## Description

We should consider skipping the Azure Static Web App step when there are missing secrets. PRs from a forked branch are not working since they don't have access to the secrets, which is blocking merging

We can depend on the Netlify deployment for checking the preview


msankaran0712 (approved):

ryan-henness-trimble:
Fixed placement here: https://github.com/trimble-oss/modus-web-components/pull/1383

Issue #1301, opened 2023-05-17
TP-7019: modus-text-input and modus-number-input components should have ability to right align the value


kmamidipalli:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have read the [contributing guidelines](https://github.com/trimble-oss/modus-web-components/blob/main/CONTRIBUTING.md)

### Proposal

![image](https://github.com/trimble-oss/modus-web-components/assets/40302378/cf57adc9-b2ac-44ff-8d19-85967ce55003)

Align the text to the right of the modus-text-input and modus-number-input components.

### Motivation and context

monetary amounts are better read right-aligned when using two decimal places, especially when using several fields in a single-column form

### Which npm package?

@trimble-oss/modus-web-components

### Priority

Medium

### What is the product name the feature is requested for?

e-Builder

### What is the team name the product belongs to?

e-Builder

### What is the division name the team belongs to?

e-Builder

msankaran0712:
It is possible to set text-align to right for the HTML input. @egunther39 @enowak1031 @soniakaukonen Is this a valid use case ? If yes, we need to add a configuration.

msankaran0712:
@kmamidipalli is your team willing to contribute?

msankaran0712:
@kmamidipalli The issue has been fixed and closed.

kmamidipalli:
> @kmamidipalli The issue has been fixed and closed.

Alright, Thanks @msankaran0712

PR #1406, opened 2023-06-08
add properties to align right on text and number inputs

acomsto:
## Description
Add ability to configure the modus-text-input and modus-number-input text alignment to be left or right.

To address this problem I added properties to align right on modus-text-input and modus-number-input, and default this property to be left aligned.  This also allows for the inclusion of center alignment at some point in the future if needed.  The docs and control for modus-text-input and modus-number-input have been updated, as well as the tests.

References #1301 

## Type of change

- [x] Documentation update

## How Has This Been Tested?

Manually tested as well as adding jest tests to check for the addition of the new classes and styling.

## Checklist

### DIFF
--- a/stencil-workspace/src/components/modus-number-input/modus-number-input.tsx
+++ b/stencil-workspace/src/components/modus-number-input/modus-number-input.tsx
@@ -43,6 +43,9 @@ export class ModusNumberInput {
   /** (optional) The input's step. */
   @Prop() step: number;
 
+  /** (optional) Set text alignment for the number input. */
+  @Prop() textAlign: 'left' | 'right' = 'left';
+
   /** (optional) The input's valid state text. */
   @Prop() validText: string;
 
@@ -78,6 +81,7 @@ export class ModusNumberInput {
     const inputContainerClassName = `input-container ${
       this.errorText ? 'error' : this.validText ? 'valid' : ''
     } ${this.classBySize.get(this.size)}`;
+    const textAlign = `text-align-${this.textAlign}`;
 
     return (
       <div
@@ -99,6 +103,7 @@ export class ModusNumberInput {
         ) : null}
         <div class={inputContainerClassName}>
           <input
+            class={`${textAlign}`}
             disabled={this.disabled}
             max={this.maxValue}
             min={this.minValue}
--- a/stencil-workspace/src/components/modus-text-input/modus-text-input.scss
+++ b/stencil-workspace/src/components/modus-text-input/modus-text-input.scss
@@ -67,6 +67,10 @@
         padding-right: 0;
       }
 
+      &.text-align-right {
+        text-align: right;
+      }
+
       &::placeholder {
         color: $modus-input-hint-text-color;
       }
--- a/stencil-workspace/src/components/modus-text-input/modus-text-input.tsx
+++ b/stencil-workspace/src/components/modus-text-input/modus-text-input.tsx
@@ -59,6 +59,9 @@ export class ModusTextInput {
   /** (optional) The input's size. */
   @Prop() size: 'medium' | 'large' = 'medium';
 
+  /** (optional) Set text alignment for the input. */
+  @Prop() textAlign: 'left' | 'right' = 'left';
+
   /** (optional) The input's type. */
   @Prop() type: 'text' | 'password' = 'text';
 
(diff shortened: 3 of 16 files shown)


msankaran0712 (changes requested):
All good except a padding issue. Please fix the unnecessary right padding when the text is right aligned. 
If you see when the clear icon is added, the unwanted space between the text and the icon is removed which looks correct. 
So, you have to set the `padding-right` to zero for the text by default when it is right aligned.

cc @ryan-henness-trimble 

msedge_YoVXOhQKVl

stencil-workspace/src/components/modus-text-input/modus-text-input.tsx
```diff
@@ -59,6 +59,9 @@ export class ModusTextInput {
   /** (optional) The input's size. */
   @Prop() size: 'medium' | 'large' = 'medium';
 
+  /** (optional) Set text alignment for the input. */
```
ryan-henness-trimble, line 62: Consistency consideration: Change this comment to `The input's text alignment.`

stencil-workspace/src/components/modus-text-input/modus-text-input.tsx
```diff
@@ -116,6 +119,10 @@ export class ModusTextInput {
     const isPassword = this.type === 'password';
     const showPasswordToggle = !!(this.includePasswordTextToggle && isPassword && this.value?.length);
     const showClearIcon = (isPassword && !this.includePasswordTextToggle) || !isPassword;
+    const textAlign = `text-align-${this.textAlign}`;
+    const inputClassName = `${this.includeSearchIcon ? 'has-left-icon' : ''} ${
```
ryan-henness-trimble, line 123: Few things here:

We could think about making the name plural here,`inputClassNames`. I think it would make sense to update the constant above it too, to `classNames` or `containerClassNames`. I think leaving the name singular here has been an oversight in past changes

The `textAlign` const  could be built into the string interpolation for `inputClassNames`

Since it's requiring more logic to build out these class names, and we're unnecessarily adding an empty string when the classes don't apply, we should also think about adding a function to build this string out instead, it might make it easier to read:
```ts
const buildTextInputClassNames = (): string => {
  const classNames = [];
 
 if (this.includeSearchIcon) {
    classNames.push('has-left-icon');
  }

  if (this.clearable) {
    classNames.push('has-right-icon');
  }

  classNames.push(`text-align-${this.textAlign}`);
  
  return classNames.join(' ');
}
```

stencil-workspace/src/components/modus-text-input/modus-text-input.scss
```diff
@@ -67,6 +67,11 @@
         padding-right: 0;
       }
 
+      &.text-align-right {
+        padding-right: 0;
```
ryan-henness-trimble, line 71: Do we want to make this `$rem-8px` instead of adding the `hidden` class?

stencil-workspace/src/components/modus-text-input/modus-text-input.tsx
ryan-henness-trimble: If we make this function, we could call it directly in the JSX if we want

msankaran0712 (changes requested):
@egunther39, Can you please review the right-aligned text input? https://deploy-preview-1406--modus-webcomponents.netlify.app/?path=/story/user-inputs-text-input--default&args=clearable:true;textAlign:right

egunther39 (approved):

stencil-workspace/src/components/modus-number-input/modus-number-input.tsx
```diff
@@ -99,6 +103,7 @@ export class ModusNumberInput {
         ) : null}
         <div class={inputContainerClassName}>
           <input
+            class={`${textAlign}`}
```
ryan-henness-trimble, line 106: We should rename this `textAlign` to `textAlignClassName`, and we can pass it to the input without the interpolated string, like:
```ts
class={textAlignClassName}
```

stencil-workspace/src/index.html
```diff
@@ -17,5 +17,19 @@
   </head>
   <body>
     <!-- Test components here -->
+
+    <modus-text-input
```
ryan-henness-trimble, line 21: Once we're ready to merge, we should revert this file back to the original state

stencil-workspace/src/components/modus-text-input/modus-text-input.tsx
```diff
@@ -163,14 +180,11 @@ export class ModusTextInput {
...
           )}
-          {showClearIcon &&
-            (this.clearable && !this.readOnly && !!this.value ? (
-              <span class="icons clear" role="button" aria-label="Clear entry">
-                <IconClose onClick={() => this.handleClear()} size="16" />
-              </span>
-            ) : (
-              <span class="icons"></span>
-            ))}
+          {showClearIcon && this.clearable && !this.readOnly && !!this.value && (
```
ryan-henness-trimble, line 183: We should fix this logic check here `showClearIcon && this.clearable && !this.readOnly && !!this.value`

it seems like it should read just like `showClearIcon && <...element>`

ryan-henness-trimble (approved):

Issue #1334, opened 2023-05-24
[BUG] Modus Select Component - State issues


msankaran0712:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have validated any HTML to avoid common problems
- [X] I have read the [contributing guidelines](https://github.com/trimble-oss/modus-web-components/blob/main/CONTRIBUTING.md)

### Describe the issue

- Select dropdown icon on right isn't the correct one
- input options are not correct font
- Focus state doesn't have blue bottom border
- Also, the active class is applied to any option that is hovered over creating confusion about whether an item is hovered or selected.
- Can we have a placeholder text for this control? Thanks!

Split from #559 

### What operating system(s) are you seeing the problem on?

Windows

### Which npm package?

@trimble-oss/modus-web-components

### Priority

High

### What product/project are you using Modus Components for?

Modus

### What is your team/division name?

Modus

sarafarhat13:
Any updates on the placeholder text? The web component still doesn't have a placeholder text.

Issue #559, opened 2022-07-12
[BUG] Select Component UI issues


coliff:
https://modus-web-components.trimble.com/?path=/docs/user-inputs-select--default

- Select dropdown icon on right isn't the correct one
- input options are not correct font
- Focus state doesn't have blue bottom border

The component isn't a native select component which is causing accessibility issues. The dropdown isn't accessible via keyboard Enter for example.

@ryan-henness-trimble 
Could we consider using a native select instead?

sarafarhat13:
Can we have a placeholder text for this control? Thanks!

msankaran0712:
Also, the `active` class is applied to any option that is hovered over.

msankaran0712:
Added `data-table` tag because the bug should be fixed for Table pagination's page size options

HarshithaNous:
@hemantk1325 adding the below scenario with respect to keyboard accessibility

User will hit on tab key to select the first option on Select component "Select Demo1" and then hit on Enter key for the dropdown values to be displayed. Select any of the option by hitting Enter key, once the option is selected, it is expected that the dropdown should auto close. 
However this works fine when user does the action on mouse click, the same should work for action performed through keyboard.

msankaran0712:
@hemantk1325 In this issue please concentrate on only using native select, we have split the bug into multiple issues

PR #1407, opened 2023-06-09
Fix Modus Select Accessibility issues and Keyboard Navigation

hemantk1325:
## Description

div was replaced with "select native" to address the modus select issue, and the keyboard navigation problem was resolved.

References #559 ,#1334

## Type of change

## Checklist

### DIFF
--- a/stencil-workspace/src/components/modus-select/modus-select.tsx
+++ b/stencil-workspace/src/components/modus-select/modus-select.tsx
@@ -1,14 +1,5 @@
-import {
-  Component,
-  Element,
-  Event,
-  EventEmitter,
-  h, // eslint-disable-line @typescript-eslint/no-unused-vars
-  Listen,
-  Prop,
-  State,
-} from '@stencil/core';
-import { IconTriangleDown } from '../icons/icon-triangle-down';
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import { Component, Event, EventEmitter, h, JSX, Prop, State, Watch } from '@stencil/core';
 import { createGuid } from '../../utils/utils';
 
 @Component({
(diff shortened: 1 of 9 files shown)


msankaran0712 (approved):

stencil-workspace/src/components/modus-select/modus-select.tsx
```diff
@@ -58,135 +40,99 @@ export class ModusSelect {
...
   }
 
-  @Listen('keydown')
-  elementKeydownHandler(event: KeyboardEvent): void {
-    if (!this.visible || this.el.shadowRoot.activeElement.tagName !== 'BUTTON') {
-      return;
-    }
+  handleItemSelect(option: unknown): void {
+    this.valueChange.emit(option);
+    this.internalValue = option;
```
msankaran0712, line 65: @hemantk1325 I would update the internal state only when the `value` changes, not on `valueChange` because it can lead to a stale state if the consumer decides not to update the `value`. Ideally, it should behave like https://codesandbox.io/s/happy-sunset-t6c8tg?file=/demo.tsx, where the consumer gets complete control over the `value`, and the input should always render the `value`.

@ryan-henness-trimble, please correct me if you think otherwise

msankaran0712: @hemantk1325 that means we should update the storybook examples to update the `value` inside `valueChange`

msankaran0712 (changes requested):

ryan-henness-trimble (approved):

msankaran0712 (approved):

coliff (approved):

coliff:
This is great - a huge improvement - thanks!

Issue #1044, opened 2023-02-21
[BUG] Accessibility Issue with Text Input and Number Input


msankaran0712:
- Form elements must have labels.
- Ensures ARIA attributes are allowed for an element's role

**To Reproduce**
Goto Accessibility Tab in [Modus Number Input](https://modus-web-components.trimble.com/?path=/story/user-inputs-number-input--default)
Goto Accessibility Tab in [Modus Text Input](https://modus-web-components.trimble.com/?path=/story/user-inputs-text-input--default)

**Expected behavior**
Fix accessibility issues

**Screenshots**
![image](https://user-images.githubusercontent.com/50131391/220281946-a821a222-68b9-4edc-ac0b-2ac5f3ab6b3c.png)

UPDATE:
This issue is partially fixed as user can set an aria-label, however this shouldn't be necessary. Setting a label should automatically link to the form and is a better solution.
![image](https://github.com/trimble-oss/modus-web-components/assets/1212885/4b23d839-923e-4780-899d-a3d9626b19ff)

Evaluate all inputs for labels and work to have them utilize the `for` attribute correctly.

PR #1611, opened 2023-08-04
Fix Accessibility Errors on the User Input Components

msankaran0712:
The `aria-` attributes moved from the `div` to the `input` element. 

Note: 
Did not fix the `aria-placeholder` on the Number Input Component as it is not clear whether it is okay to remove the attribute.

References #1044

## Type of change

- [ ] Documentation update

## Checklist

### DIFF
--- a/stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
+++ b/stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
@@ -229,12 +229,9 @@ export class ModusDateInput {
             this.size
           )}`}>
           <input
-            aria-disabled={this.disabled ? 'true' : undefined}
             aria-invalid={!!this.errorText}
             aria-label={this.ariaLabel}
-            aria-readonly={this.readOnly}
-            aria-required={this.required}
-            aria-placeholder={this.placeholder}
+            aria-required={this.required?.toString()}
             autofocus={this.autoFocusInput}
             class={{ 'has-right-icon': this.showCalendarIcon }}
             disabled={this.disabled}
--- a/stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
+++ b/stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
@@ -217,13 +217,7 @@ export class ModusDateInput {
   render() {
     const className = `modus-date-input ${this.disabled ? 'disabled' : ''}`;
     return (
-      <div
-        aria-disabled={this.disabled ? 'true' : undefined}
-        aria-invalid={!!this.errorText}
-        aria-label={this.ariaLabel}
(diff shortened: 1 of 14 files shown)


coliff (commented):
I've spotted some issues:

- `aria-readonly` isn't being set correctly. On this URL: https://deploy-preview-1611--modus-webcomponents.netlify.app/?path=/story/user-inputs-text-input--default&args=ariaLabel:;readOnly:true;required:true the element receives `aria-readonly` with no value which means it is NOT readonly. For the element to be ready-only it needs to be `aria-readonly="true"` (as the default is false). Overall, I think it maybe better to not use ARIA label - you can simply set the readonly HTML attribute and that will be announced to   assistive tools.

- I don't see an issue with the `aria-disabled` implementation, but f the disabled attribute is being added too then the aria-disabled label isn't needed.

- `aria-required` is not set correctly. It should be `aria-required="true"`

- `aria-placeholder` is not set correctly. Suggest to use standard placeholder HTML attribute instead.

- This might be a Storybook bug, but if I change the input type to password I see a serious a11y warning 'aria-label attribute cannot be used on a div with no valid role attribute.' It looks like `aria-label` is on the div but with no label added - can it be removed altogether as to not cause potential issues?

msankaran0712:
> * `required`

@coliff I have removed redundant `aria-disabled`, `aria-placeholder` and `aria-readonly` as we have the corresponding attribute to represent the state and for password text input, added a button role to the div.

coliff:
Not working right now - the generated HTML code for the input is:

```html
<input aria-disabled="true" aria-label="Text Input" aria-readonly="" aria-required="" aria-placeholder="" class="text-align-left" disabled="" inputmode="" maxlength="100" minlength="0" placeholder="" readonly="" tabindex="0" type="password" autofocus="">
```

Here's the a11y panel in DevTools
image

ryan-henness-trimble (approved):

msankaran0712:
> Not working right now - the generated HTML code for the input when I turn on disabled/readonly and required is:
> 
> ```
> <input 
> aria-disabled="true" 
> aria-label="Text Input" 
> aria-readonly="" 
> aria-required="" 
> aria-placeholder="" 
> class="text-align-left" 
> disabled="" 
> inputmode="" 
> maxlength="100" 
> minlength="0" 
> placeholder="" 
> readonly="" 
> tabindex="0" 
> type="password" 
> autofocus="">
> ```
> 
> Here's the a11y panel in DevTools image

@coliff Thats weird because Azure deploy does not have the latest fix. Please check the Netlify preview once https://deploy-preview-1611--modus-webcomponents.netlify.app/?path=/story/user-inputs-text-input--default&args=disabled:true;readOnly:true;required:true;type:password

coliff (approved):

Issue #1729, opened 2023-09-22
Modus Table - Add row actions 


msankaran0712:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have read the [Modus style guidelines](https://modus.trimble.com/components/web/introduction/)

### Proposal

We need row actions for Modus Table component as shown below, and the Figma prototype can be found in this [link](https://www.figma.com/file/7yC5vYgdbzqaEJDr566xAL/SUBMISSION-V1---Work-in-Progress?node-id=0%3A1&mode=dev)

![image](https://github.com/trimble-oss/modus-web-components/assets/50131391/046ef19c-bcef-449e-bb65-1c07510be986)

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

High

### What product/project are you using Modus Components for ?

modus

### What is your team/division name ?

modus

### Are you willing to contribute ?

None

msankaran0712:
Maha to check with Matt and Modus team to revise the guidelines for row actions in Modus Styleguide

msankaran0712:
From a discussion with Matt Brinkman of the Tables Tiger team, it was decided that the new Table component should display an overflow icon in the first cell of each row for row actions, prototype can be found in this [Figma link](https://www.figma.com/file/7yC5vYgdbzqaEJDr566xAL/SUBMISSION-V1---Work-in-Progress?node-id=0%3A1&mode=dev)
![image](https://github.com/trimble-oss/modus-web-components/assets/50131391/1dc84d12-01a4-468d-b38f-ef866a50cb68)

cc @enowak1031 @egunther39

msankaran0712:
Leonardo Rodrigues is working on this issue

enowak1031:
@msankaran0712 Moving to Sprint 73. Let's get an update.

PR #1736, opened 2023-09-27
1729 Modus Table - Add row actions

leosole-trimble:
## Description

Added row actions to Modus Table

References #1729 

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

I added the component with the new features to the index.html file and added new tests

## Checklist


stencil-workspace/src/components/modus-table/modus-table.tsx
```diff
@@ -584,6 +638,34 @@ export class ModusTable {
...
+    });
+    return (
+      <div
+        id={`dropdown-${id}`}
+        class="dropdownMenu"
+        style={{ top: `${this.dropdownY}px`, left: `${this.dropdownX}px`, position: 'absolute' }}>
+        {this.table.getRowModel()?.rows.filter((row) => row.id == id).map((row) => (
+            <div class="list-container" id={row.id}>
+              <div class="items-container">
+                {this.overflowMenuActions.filter(action => !row.original["_excludedActions"]?.includes(action._id)).map((action) => (
```
leosole-trimble, line 653: There's probably a better way to do that, but I couldn't find how to add the `_excludedActions` property to the row itself, only to the original. The goal was to enable the removal of certain actions from a row (like the row represents a protected element that can't be deleted for example)

stencil-workspace/src/components/modus-table/modus-table.tsx
```diff
@@ -167,6 +169,34 @@ export class ModusTable {
...
+        ...this.toolbarOptions.columnsVisibility.requiredColumns,
+        ...this.frozenColumns,
+      ];
+    }
+  }
+
+  /** (Optional) Dropdown menu with actions for each row. */
+  @Prop() overflowMenuActions?: ModusTableRowAction[] = [];
+  @Watch('overflowMenuActions') onChangeOfOverflowMenuActions() {
+    if (this.overflowMenuActions.length > 0) {
```
cjwinsor, line 189: This will error when `this.overflowMenuActions` is null

cjwinsor: I mention this as the deployed example site is showing errors in the console related to this line.

stencil-workspace/storybook/stories/components/modus-table/modus-table-storybook-docs.mdx
```diff
@@ -441,35 +467,39 @@ Users can use keyboard navigation to perform different actions.
...
+| `summaryRow`           | `summary-row`             | (Optional) To display summary row.                                                                                                  | `boolean`                              | `false`                                                   |
+| `toolbar`              | `toolbar`                 | (Optional) To display a toolbar for the table.                                                                                      | `boolean`                              | `false`                                                   |
+| `toolbarOptions`       | --                        | (Optional) To display a toolbar, which allows access to table operations like hiding columns.                                       | `ModusTableToolbarOptions`             | `null`                                                    |
+| `rowActions`           | --                        | (Optional) Actions that can be performed on each row. A maximum of 4 icons will be shown, including overflow menu and expand icons. | `ModusTableRowAction[]`                | `[]`                                                      |
```
cjwinsor, line 491: May be good to sort the new properties into alphabetical order here, vs adding to the end

stencil-workspace/storybook/stories/components/modus-table/modus-table-storybook-docs.mdx
```diff
@@ -441,35 +467,39 @@ Users can use keyboard navigation to perform different actions.
...
-| `cellLinkClick`      | Emits the link that was clicked            | `CustomEvent<ModusTableCellLink>` |
-| `rowSelectionChange` | Event details contains the row(s) selected | `CustomEvent<unknown>`            |
-| `sortChange`         | Emits event on sort change                 | `CustomEvent<ColumnSort[]>`       |
+| Event                         | Description                                                 | Type                                         |
+| ----------------------------- | ----------------------------------------------------------- | -------------------------------------------- |
+| `cellLinkClick`               | Emits the link that was clicked                             | `CustomEvent<ModusTableCellLink>`            |
+| `rowSelectionChange`          | Event details contains the row(s) selected                  | `CustomEvent<unknown>`                       |
+| `sortChange`                  | Emits event on sort change                                  | `CustomEvent<ColumnSort[]>`                  |
+| `rowActionClick`              | An event that fires when a row action is clicked.           | `CustomEvent<ModusTableRowActionClickEvent>` |
```
cjwinsor, line 501: May be good to sort the new properties into alphabetical order here, vs adding to the end

cjwinsor:
I haven't seen a finalized figma on how this should be visually, so some of these are just thoughts.

## For @leosole-trimble:

### MouseEvent Handling:
- The method `renderDropdownMenu` adds a `click` event handler using the `x` and `y` properties from `MouseEvent`. They are aliases for `clientX` and `clientY`, which provide coordinates relative to the viewport. However, `pageX` and `pageY` provide coordinates relative to the entire document, factoring in scroll offset. This distinction might not have been clear in the storybook since it's presented within an iframe.

### Dropdown Behavior:
- Consider making the menu appear relative to the clicked element, rather than the click location for a more standardized and clean user experience.

### DOM Rendering:
- Instead of pre-rendering, perhaps generate the list on demand to avoid leaving an unused `<div id="dropdown-null" class="dropdownMenu" style="top: 0px; left: 0px; position: absolute;"></div>` in the DOM. Consider handling this within the `modus-table-row-actions` component.

### Component Responsibility:
- Should we introduce a `modus-table-overflow-menu` component to better separate responsibilities?

### Styling
- Could we avoid passing `isChecked` down if we updated the css to properly account for how we want a `row-action` to be styled?

## For @msankaran0712:

### Cell Configuration:
- Should we allow row actions and overflow menus in the same cell simultaneously?
- Is this intended to be a distinct column type, or can row actions and overflow menus coexist with content within the same cell?
- Should we be nesting `modus-table-row-actions` under `parts > cell` or `parts > row` vs the root of `parts`?

### Component Usage:
- Direct use of `modus-dropdown` might lead to redundancy, but employing `modus-list` seems viable, especially if rendered on demand. However, I don't feel the styling of the list has been perfected yet.

### Design Considerations:
- Should the menu always occupy the first cell? There could be use cases that desire it to be the last column.
- For sub-rows, should action items be nested along with the content or aligned at the start of the cell?

### Prop Naming and Structure:
- Both `rowActions` and `overflowMenuActions` use the `ModusTableRowAction` type, but not every property (like `icon`) is respected.
- Consider restructuring `rowActions` and `overflowMenu` to include an `actions` property to support future extension.
- The `_id` in the actions array might be more semantically named as `id` or `key`.
- Evaluate if nesting `text` and `icon` under `display` is necessary. Also, `text` vs `label` here.

### Terminology:
- Is `actions` the best term, or should we use `action-items` for consistency, or just `items`?

### Template Management:
- I'm not a fan of an exclude list. Perhaps, allow row-level template overrides or offer format functions for consumers to tailor row actions or overflow items dynamically.  Alternatively, actions could be a function or an array.. and when its a function, similar processing logic would apply except now you'd have to return the set of actions you want to include.

```js
table.rowActions = {
    actions: [{
        id: 'edit',
        label: 'Edit',
        icon: 'edit',
        isVisible: (row) => row.mode !== 'readOnly'
    },{
        id: 'delete',
        label: 'Delete',
        icon: 'delete',
        isVisible: (row) => row.mode !== 'readOnly'
    }]
};
```
and 
```jsx
// Assuming row is passed in via props
// Should have checks for isVisible is function or boolean, 
// perhaps a wrapping function is called to handle that before calling users function.
{ props.actions.map((action) => (
    action.isVisible(props.row) && <div
        class={`row-action ${props.isChecked ? 'row-selected' : ''}`}
        onClick={() => props.onRowActionClick(action.id, props.rowId)}>
        <IconMap icon={action.display.icon} size="16" /> 
    </div>
)}
```
And similar or overflow menu.  This added complexity is why I am thinking that the overflow and row actions logic gets broken up.  

This is as far as I can take my analysis today.

msankaran0712:
Thank you for the thorough review, @cjwinsor. It was truly helpful.👍🏼 

Hey @leosole-trimble, sorry for the delay. Please take a look at my comments below.

- The overflow menu only enables users to perform more actions when the row should have only a maximum of 4 icons in the user interface. So, I agree with @cjwinsor on keeping overflow menu items inside the `rowActions` property, we don't need separate configurations for the actions and the overflow menu. I would suggest the below structure for that configuration, and you can see that `showInMenu` decides whether an action should be shown as an icon or inside the menu.
```js
this.table.rowActions = [
  {
    id: 'delete', 
    icon: 'trash',
    label: 'Delete',
    onClick: (row) => {
      console.log('Delete row', row);
    },
    isVisible: (row) => {
      return row.id !== 1;
    },
  },
  {
    id: 'edit', 
    icon: 'edit',
    label: 'Edit',
    onClick: (row) => {
      console.log('Edit row', row);
    },
    isVisible: true,
    showInMenu: true, // Icons will be not shown in the menu
  },
]
```
- I would recommend using the `modus-dropdown` with `modus-button`(icon-only variant) and `modus-list` so it can take care of rendering the menu items, avoiding the code for handling opening/closing the menu, and positioning it.  The list is not borderless, but we can fix that part in this issue.
- I agree with @cjwinsor about moving the row actions to a shadowless component(`modus-table-row-actions.tsx`), keeping the code clean and maintainable, and it should reside under the `parts -> row` folder.
- @cjwinsor, row actions are expected to co-exist with the cell text, aligned to the sub-level. The main reason for keeping it that way is to prevent row actions from getting hidden while scrolling and maybe also to keep it simple. I am sure the Tables Tiger team must have considered having a separate action column. If there is any use case, we can talk to Matt Brinkman. 
- As part of the Inline editing work, I have done some code refactoring that also handles the passing of `isChecked` down to the cells. And I would place row actions component [here](https://github.com/trimble-oss/modus-web-components/blob/face7fd31ff7d460dc2938804e03ca4ba1b180f0/stencil-workspace/src/components/modus-table/parts/cell/modus-table-cell-main/modus-table-cell-main.tsx#L125)

msankaran0712:
@leosole-trimble @cjwinsor Please let me know if there is anything not addressed.

cjwinsor:
@msankaran0712 On the use of `modus-dropdown`, the cell would build the menu on initial render for each row, where using the `modus-list` would enable us to defer the rendering of the menu until they click on the action.  I'm not sure if its premature optimization, but it would reduce the size of the dom tree.  

Also the addition of `onClick` as a property of the action seems like a departure from the existing pattern where the component fires a generic event.  I'm fine either way, just noting the change.

Did you have thoughts on how `showInMenu` works in relation to the max of 4 items?  If we have 6 items but only one sets `showInMenu` as true, for example.  If no icon is defined then the button is currently empty.  Given that, you could say that if an action is given an icon then it will show as an action button up till the first 4 then the rest go into overflow.  If no icon is provided, it goes directly to overflow.  This assumes we don't want the `label` to be used in place of the `icon`.

msankaran0712:
> @msankaran0712 On the use of `modus-dropdown`, the cell would build the menu on initial render for each row, where using the `modus-list` would enable us to defer the rendering of the menu until they click on the action. I'm not sure if its premature optimization, but it would reduce the size of the dom tree.
> 
> Also the addition of `onClick` as a property of the action seems like a departure from the existing pattern where the component fires a generic event. I'm fine either way, just noting the change.
> 
> Did you have thoughts on how `showInMenu` works in relation to the max of 4 items? If we have 6 items but only one sets `showInMenu` as true, for example. If no icon is defined then the button is currently empty. Given that, you could say that if an action is given an icon, then it will show as an action button up till the first 4 then the rest go into overflow. If no icon is provided, it goes directly to overflow. This assumes we don't want the `label` to be used instead of the `icon`.

Thanks @cjwinsor, you are right about the Modus Dropdown, although we have pagination and lazy loading, and not many row actions can be there in the menu, considering how the Modus dropdown handles list items may not be so ideal. I was trying to reuse the Dropdown logic only if it could take an object array and render it dynamically.

Yeah, I thought so too, I was speculating about the idea of adding the handler inside the config, am fine with events. 

Regarding the limitation of icons, I think there might be use cases where the user is interested only in the menu, not the visible icons. We can add it to the documentation about how many icons are permitted, and we internally can stop rendering the icons more than 4.

cjwinsor:
@leosole-trimble A follow-up and @msankaran0712 and I had a discussion.  We can have the structure be
```js
this.table.rowActions = [
  {
    id: 'delete', 
    icon: 'trash',
    label: 'Delete',
    isVisible: (row) => {
      return row.id !== 1;
    },
  },
  {
    id: 'edit', 
    icon: 'edit',
    label: 'Edit',
    isVisible: true,
  },
]
```

And it will function with the following rules: 
1. No more than 4 total icons (included in this count is the expandable icon and the menu icon itself)
2. After the max of 4 total icons, the remaining items go into the overflow menu
3. No decision was made around handling an action that doesn't have an icon defined, so for now it can be an empty button?
4. There is no property at this time to specify that an item goes into the overflow menu, that is all handled by the control logic.

For clarity, here are examples
- If the row is showing the expandable icon, then 3 actions can be shown before the overflow menu is introduced
- If the row is showing the expanding icon and the user has 4 actions, you'll see the expandable icon, 2 action icons, and the menu icon, with the menu containing the last 2 menu options.

Aside from that, we discussed not using the `modus-dropdown` but we should probably still use the `modus-list`.  The menu should function as if it was a drop down though. We aren't suggesting the use of `modus-dropdown` due to the number of DOM elements it would create doing so, and its a better strategy to conditionally build that menu as needed with the items it needs.

Overall this simplifies the approach and avoids over engineering at this time while still allowing for future additions that won't cause breaking changes.  So think of it as an array of actions, and the control will control how it overflows to maintain a max of 4 total icons displayed at a time.

msankaran0712:
@leosole-trimble, I would recommend to refer `Modus Dropdown` code for how it handles showing & hiding the menu, excluding the slot option, it does not use x and y.

leosole-trimble:
Questions:
1) I'm not really using the `modus-dropdown` component here, just `div`s. Should I use `modus-list` anyway?
2) Should the menu be inside the cell? I tried it, but can't find a way to make the menu appear over the other rows.

leosole-trimble:
I managed to do it passing a ref to the clicked element. Not sure if this is the right way

msankaran0712:
Closing the PR as there is a new https://github.com/trimble-oss/modus-web-components/pull/1784 with latest

Issue #1075, opened 2023-03-01
Add `min` and `max` date for Date picker in Web components


msankaran0712:

msankaran0712:
[PR ](https://github.com/trimble-oss/modus-web-components/pull/1738#pullrequestreview-1649511823) has been submitted

msankaran0712:
@yevhenkravets @cjwinsor, I have copied here your requirements from the [PR](https://github.com/trimble-oss/modus-web-components/pull/1738#pullrequestreview-1649511823), Please feel free to add comments

- Update the modus web component with a parameter to support a minimum and/or maximum value
- Validate input validation is working (max date value, min date value)
- Min error message = “Select a date after [min value - 1 day]”
- Max error message = “Select a date before [max value + 1 day]”
- Disabled days in the calendar widget must show a tooltip when on hover/focus
- Tooltip Message = min/max error message

cc @egunther39 @enowak1031 @soniakaukonen @coliff

msankaran0712:
@yevhenkravets PR merged and closed

PR #1738, opened 2023-09-28
EPD-11512 Issue 1075/date picker min max

yevhenkravets:
## Description

- Update the modus web component with a parameter to support a minimum and/or maximum value 
- Validate input validation is working (max date value, min date value)
- Min error message = “Select a date after [min value - 1 day]”
- Max error message = “Select a date before [max value + 1 day]”
- Disabled days in the calendar widget must show a tooltip when on hover/focus
- Tooltip Message = min/max error message

References #1075 

## Type of change

- [x] Documentation update

## How Has This Been Tested?

Provide min and/or max attr value for the input in the storybook and check manual typing or pasting. Also, check with the date picker.
Added related integration tests for the input and picker components

## Checklist


stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
```diff
@@ -157,12 +167,24 @@ export class ModusDateInput {
     this._dateInput.focus();
   }
 
+  /** Validate the input. */
+  @Method()
+  async validate(): Promise<void> {
+    this.validateInput(this._dateDisplay);
+  }
```
yevhenkravets, line 174: public method to validate the input. For example, if the date range is fine and you clear the error message, but one of the fields is out of the min-max range you call it to check the validation on the input itself.

stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
```diff
@@ -157,12 +167,24 @@ export class ModusDateInput {
...
+    const messages = this.getMinMaxValidationMessage();
+
     this.calendarIconClicked.emit({
       value: this.value,
       type: this.type,
       inputString: this._dateDisplay,
+      min: this._formatter.parseIsoToDate(this.min),
+      max: this._formatter.parseIsoToDate(this.max),
+      minMessage: messages.min,
+      maxMessage: messages.max,
```
yevhenkravets, line 187: The calendar doesn't know anything about the format, but the tooltip message of the disabled date should be the same as the input one. Send them to the calendar with the min and max values.

stencil-workspace/src/components/modus-date-input/utils/modus-date-input.formatter.tsx
```diff
@@ -1,6 +1,6 @@
 import { TokenFormatting, token, Tokens, tokenType, TokenParser } from './modus-date-input.tokens';
 
-export const ISO_DATE_FORMAT = /^(\d{4})-(1[0-2]|0?[1-9])-(3[01]|0?[1-9]|[12][0-9])$/;
+export const ISO_DATE_FORMAT = /^(\d{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])/;
```
yevhenkravets, line 3: fixed the regex. Zeros are required for the ISO format if it 1-9 value.
Also, the ISO format could have other parts of the date so it doesn't end by the day value. But, we need only year, month, and day.
https://www.w3.org/TR/NOTE-datetime

stencil-workspace/src/components/modus-date-picker/utils/modus-date-picker.state.tsx
```diff
@@ -24,6 +24,7 @@ export default class ModusDatePickerState {
 
       this.element.value = `${year}-${month}-${date}`;
       this.element.focusInput();
+      this.element.validate();
       this.date = val;
     }
```
yevhenkravets, line 29: if the input has an error message but you choose a valid date in the calendar then we need to validate the input immediately.

stencil-workspace/storybook/stories/components/modus-date-picker/modus-date-picker.stories.tsx
```diff
@@ -284,3 +308,58 @@ DateRange.args = {
...
+DefaultWithPicker.args = {
+  ...defaultArgs,
+  showCalendarIcon: true,
+  min: '2022-12-02',
+  max: '2022-12-30',
+  format: 'mmm dd, yyyy',
+  allowedCharsRegex: '.',
+  helperText: 'mmm dd, yyyy',
+  label: 'Single Date',
+};
```
yevhenkravets, line 365: Added one more template with a picker and default min and max values.

stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
```diff
@@ -157,12 +167,20 @@ export class ModusDateInput {
     this._dateInput.focus();
   }
 
+  /** Validate the input. */
+  @Method()
+  async validate(): Promise<void> {
+    this.validateInput(this._dateDisplay);
```
msankaran0712, line 173: why are we passing the local variable `this._dateDisplay` ?

msankaran0712: Ahh, okay, an existing code was doing it. I think it is safe to not pass the argument.

stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
```diff
@@ -157,12 +167,20 @@ export class ModusDateInput {
...
+  }
+
   /** Handlers */
   handleCalendarClick(): void {
     this.calendarIconClicked.emit({
       value: this.value,
       type: this.type,
       inputString: this._dateDisplay,
+      min: this._formatter.parseIsoToDate(this.min),
```
msankaran0712, line 182: I would suggest not parsing to Date objects because of timezone differences when emitting events, and I think we don't have to pass min and max because the Date picker has a state object that contains a reference to the date input node, we can directly consume date input properties in that. Passing the type only matters here.

stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
```diff
@@ -252,10 +268,48 @@ export class ModusDateInput {
...
+    } else if (max && max < value) {
+      this.errorText = messages.max
+    } else {
+      this.clearValidation();
+    }
+  }
+
+  private getMinMaxValidationMessage(): { min: string; max: string } {
+    const result = { min: '', max: '' };
+    const min = this._formatter.parseIsoToDate(this.min);
```
msankaran0712, line 300: I think we don't have to parse the min and max again as it is done already inside `validateMinMax`.

stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
```diff
@@ -252,10 +268,48 @@ export class ModusDateInput {
...
+    const max = this._formatter.parseIsoToDate(this.max);
+
+    if (min) {
+      min.setUTCDate(min.getDate() - 1);
+      result.min = `Select a date after ${this._formatter.formatDisplayString(min.toISOString())}`;
```
msankaran0712, line 305: If possible, we could also keep the logic simple by altering the message to something like `Select a minimum date` or something, just a thought.

stencil-workspace/src/components/modus-date-input/utils/modus-date-input.formatter.tsx
msankaran0712: So this is the fix for https://github.com/trimble-oss/modus-web-components/issues/1753 @DonavanPrinsTrimble

stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
yevhenkravets: That is a text from Carlos Rodriguez.

stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
yevhenkravets: Okay, I'll take a look at how to minimize this logic.

stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
yevhenkravets: Okay, I'll try that

msankaran0712:
@yevhenkravets, I have multiple PRs. If @cjwinsor can review and approve, that would be fine too👍🏼

stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
msankaran0712: Are we working on this?

stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
msankaran0712: @yevhenkravets any updates on this one?

stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
yevhenkravets: > @yevhenkravets Are we working on this?

Yes, I'll provide changes soon

stencil-workspace/src/components.d.ts
```diff
@@ -378,11 +378,19 @@ export namespace Components {
...
-        "helperText": any;
+        "helperText": string;
         /**
           * (optional) The input's label.
          */
         "label": string;
+        /**
+          * (optional) The maximum date allowed. The date is formatted according to ISO8601 'yyyy-mm-dd'.
+         */
+        "max": string;
```
cjwinsor, line 389: Could we accept Date objects for `max` and `min`?

yevhenkravets: We use the ISO8601 standard here. Date objects could cause a timezone issue.

stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
```diff
@@ -252,10 +265,32 @@ export class ModusDateInput {
...
+  }
+
+  private validateMinMax(): void {
+    const min = this._formatter.parseIsoToDate(this.min);
+    const max = this._formatter.parseIsoToDate(this.max);
+    const value = this._formatter.parseIsoToDate(this.value);
+
+    if (min && min > value) {
+      min.setUTCDate(min.getDate() - 1);
+      this.errorText = `Select a date after ${this._formatter.formatDisplayString(min.toISOString())}`;
```
msankaran0712, line 287: if `min` variable is not null, it means `this.min` is a valid ISO String, so we can write `this._formatter.formatDisplayString(this.min)`

stencil-workspace/src/components/modus-date-picker/modus-date-picker.tsx
```diff
@@ -157,15 +163,32 @@ export class ModusDatePicker {
...
-    this._dateInputs[currentDateOpen].setDate(date);
+    this._currentInput.setDate(date);
     this.toggleCalendar(false);
   }
 
   showYearChange(show = true) {
     this._showYearArrows = show;
   }
 
+  private validateMinMax(date: Date): boolean {
```
msankaran0712, line 174: since it is not used only for validation purposes, maybe rename it to `isWithinCurrentMinMax`

stencil-workspace/src/components/modus-date-picker/utils/modus-date-picker.state.tsx
```diff
@@ -24,6 +37,7 @@ export default class ModusDatePickerState {
 
       this.element.value = `${year}-${month}-${date}`;
       this.element.focusInput();
+      this.element.validate();
```
msankaran0712, line 40: Why do we validate here?

stencil-workspace/src/components/modus-date-picker/utils/modus-date-picker.state.tsx
```diff
@@ -34,18 +48,27 @@ export default class ModusDatePickerState {
 
   resetError(): void {
     this.element.errorText = null;
+    this.element.validate();
```
msankaran0712, line 51: why validation is added here?

stencil-workspace/storybook/stories/components/modus-date-picker/modus-date-picker-storybook-docs.mdx
```diff
@@ -113,6 +113,8 @@ Each character as shown in the table below can be used for representing the date
...
 | `placeholder`       | `placeholder`        | (optional) The input's placeholder text.                                                                                                                                                                          | `string`                       | `undefined`    |
 | `readOnly`          | `read-only`          | (optional) Whether the input's content is read-only                                                                                                                                                               | `boolean`                      | `undefined`    |
 | `required`          | `required`           | (optional) Whether the input is required.                                                                                                                                                                         | `boolean`                      | `undefined`    |
```
msankaran0712, line 120: We might want to expose the validate public method here.

stencil-workspace/src/components/modus-date-picker/utils/modus-date-picker.state.tsx
msankaran0712, line 40: Since we are disabling the dates outside min and max on the calendar, do we still have to validate the input?

stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
yevhenkravets: Good catch! Thank you!

stencil-workspace/src/components/modus-date-picker/modus-date-picker.tsx
yevhenkravets: Will update the name!

stencil-workspace/src/components/modus-date-picker/utils/modus-date-picker.state.tsx
yevhenkravets: Answered in another comment.

> if the input has an error message but you choose a valid date in the calendar then we need to validate the input immediately.

It was not updating the error message immediately.

stencil-workspace/src/components/modus-date-picker/utils/modus-date-picker.state.tsx
yevhenkravets: probably it doesn't make sense anymore. Let me double-check.

stencil-workspace/src/components/modus-date-input/modus-date-input.tsx
yevhenkravets: Ah, no, we cannot do that because if the min date is 10/6 we need to say `Select a date after 10/5`. Since 10/6 is the minimum **allowed** date.

stencil-workspace/src/components/modus-date-input/modus-date-input.e2e.ts
cjwinsor, line 1: Can we get boundary tests, covering when the date is before, equal, and after the min or max.  What if min and max are the same?  What about a test that has min and max set with a value in between (positive scenario).

stencil-workspace/src/components/modus-date-picker/utils/modus-date-picker.state.tsx
msankaran0712: are you trying to reset the error message because the user chose a valid date from the calendar?

stencil-workspace/src/components/modus-date-picker/utils/modus-date-picker.state.tsx
msankaran0712: I think you need that code inside `applyDateRangeRules` or `handleDateInputBlur`

stencil-workspace/src/components/modus-date-picker/utils/modus-date-picker.state.tsx
yevhenkravets: Yes, it also works. Thanks

stencil-workspace/src/components/modus-date-picker/utils/modus-date-picker.state.tsx
yevhenkravets: Updated. Take a look, please

cjwinsor (dismissed):

stencil-workspace/src/components/modus-date-picker/modus-date-picker.tsx
```diff
@@ -104,8 +110,8 @@ export class ModusDatePicker {
       this._dateInputs['start'].setError('Invalid date range');
       this._dateInputs['end'].setError();
     } else {
-      this._dateInputs['start'].resetError();
-      this._dateInputs['end'].resetError();
+      this._dateInputs['start'].validateInput();
```
msankaran0712, line 113: ideally, we should be using only the public properties and methods, here it has to be `validate()`

msankaran0712 (approved):

msankaran0712 (approved):

Issue #1753, opened 2023-10-03
Date Picker state setDate method is not ISO 8601 compliant


DonavanPrinsTrimble:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/trimble-oss/modus-web-components/blob/main/CONTRIBUTING.md)

### Describe the issue

Month and date values less than 10 are not left padded with zero, resulting in a non compliant ISO 8601 date format. 

This causes unexpected behavior when constructing a date from the element value, for example '2023-1-1'.

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

High

### What product/project are you using Modus Components for ?

Seurat

### What is your team/division name ?

Geospatial

### Are you willing to contribute ?

Yes

DonavanPrinsTrimble:
My proposed fix is to add padStart to the month and date variables before setting the element value, see below.

I will create a PR if I have access to contribute.

```
export default class ModusDatePickerState {
  ...

  setDate(val: Date): void {
    // Converting to ISO8601 'yyyy-mm-dd' format
    if (Number(val)) {
      const year = val.getFullYear();
      const month = (val.getMonth() + 1).toString().padStart(2, '0'); // Zero based number system for months
      const date = val.getDate().toString().padStart(2, '0');

      this.element.value = `${year}-${month}-${date}`;
      this.element.focusInput();
      this.date = val;
    }
  }
}
```

msankaran0712:
Hey @DonavanPrinsTrimble, thank you for reporting the issue and expressing interest in submitting a PR. Your efforts are highly appreciated. Please let us know if you have any challenges creating a PR.

msankaran0712:
@DonavanPrinsTrimble  FYI

msankaran0712:
Please look out for the next release, we are planning to do it on Friday

PR #1755, opened 2023-10-03
Fix date picker setDate iso8601 compliance

DonavanPrinsTrimble:
## Description

Add left padding to month/day values less than 10 in order to comply with ISO 8601 format.

References #1753

## Type of change

- [ X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## How Has This Been Tested?

console inspection and npm run test

## Checklist

- [ X] My code follows the style guidelines of this project
- [ X] I have performed a self-review of my own code
- [ X] My changes generate no new warnings
- [ X] New and existing unit tests pass locally with my changes
- [ X] Any dependent changes have been merged and published in downstream modules
- [ X] I have checked my code and corrected any misspellings

### DIFF
--- a/stencil-workspace/src/components/modus-date-picker/utils/modus-date-picker.state.tsx
+++ b/stencil-workspace/src/components/modus-date-picker/utils/modus-date-picker.state.tsx
@@ -19,8 +19,8 @@ export default class ModusDatePickerState {
     // Converting to ISO8601 'yyyy-mm-dd' format
     if (Number(val)) {
       const year = val.getFullYear();
-      const month = val.getMonth() + 1; // Zero based number system for months
-      const date = val.getDate();
+      const month = (val.getMonth() + 1).toString().padStart(2, '0'); // Zero based number system for months
+      const date = val.getDate().toString().padStart(2, '0');
 
       this.element.value = `${year}-${month}-${date}`;
       this.element.focusInput();


stencil-workspace/src/components/modus-date-picker/utils/modus-date-picker.state.tsx
```diff
@@ -19,8 +19,8 @@ export default class ModusDatePickerState {
     // Converting to ISO8601 'yyyy-mm-dd' format
     if (Number(val)) {
       const year = val.getFullYear();
-      const month = val.getMonth() + 1; // Zero based number system for months
-      const date = val.getDate();
+      const month = (val.getMonth() + 1).toString().padStart(2, '0'); // Zero based number system for months
+      const date = val.getDate().toString().padStart(2, '0');
```
yevhenkravets, line 23: I have seen the same issue so I fixed the Regex for the ISO8601 string.
Check it here
https://github.com/trimble-oss/modus-web-components/pull/1738/files#diff-dd46bb641b5b891f4c2ba4e47cd3057e7dd69d4d3ddc07e46af348aa7d006a7eL3

msankaran0712: I can't find it @yevhenkravets

msankaran0712 (approved):

stencil-workspace/src/components/modus-date-picker/utils/modus-date-picker.state.tsx
DonavanPrinsTrimble: @msankaran0712 This does not solve our specific case.

The reason being we are using the web components in angular and wrapping them with ControlValueAccessor. 

So in this case when you pick a date for example 2023-01-01 the value 2023-1-1 is written to the element and the parameter val passed to the handler below is formatted incorrectly and in turn emits a valueChange with an incorrectly formatted value property.

So the regular expression or formatter does not address my issue because I'm using the value property from the emitted event below.

modus-date-input.tsx
```
  /** (optional) A string representing the date entered in the input. The date is formatted according to ISO8601 'yyyy-mm-dd'. The displayed date format will differ from the 'value'. */
  @Prop({ mutable: true }) value: string;
  @Watch('value')
  handleValueChange(val: string): void {
    if (!this._isEditing) {
      this._dateDisplay = this._formatter.formatDisplayString(val);
    }

    this.valueChange.emit({
      value: val,
      type: this.type,
      inputString: this._dateDisplay,
    });
  }
```

PR #1892, opened 2023-11-16
Set background color for readonly inputs

coliff:
## Description

Set background color for readonly text input. This works good in all modern browsers, but it does use the :has selector which is relatively new. (REF: https://caniuse.com/css-has).

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

Locally with Edge v119. Tested in light and dark mode.

## Checklist

### DIFF
--- a/stencil-workspace/src/components/modus-text-input/modus-text-input.scss
+++ b/stencil-workspace/src/components/modus-text-input/modus-text-input.scss
@@ -134,6 +134,11 @@
     }
   }
 
+  .input-container:has(input[readonly]) {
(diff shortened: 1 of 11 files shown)


stencil-workspace/src/components/modus-text-input/modus-text-input.scss
```diff
@@ -134,6 +134,11 @@
     }
   }
 
+  .input-container:has(input:read-only) {
```
msankaran0712, line 137: can we do the same for all other input controls?

msankaran0712 (approved):

Issue #1930, opened 2023-12-05
Implement "Density" control on modus table


apaddock:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have read the [Modus style guidelines](https://modus.trimble.com/components/web/introduction/)

### Proposal

Relaxed = Default size: 48px/3rem in height
![Relaxed](https://github.com/trimble-oss/modus-web-components/assets/4574058/45eff661-5be4-401b-821c-14868241fc8f)

Comfortable = Condensed (Medium) size: 32px/2rem in height
![Comfortable](https://github.com/trimble-oss/modus-web-components/assets/4574058/5d7a1bc7-a1c7-445c-a469-fb28f87c70fd)

Compact = Small size: 24px/1.5 rem in height
![Compact](https://github.com/trimble-oss/modus-web-components/assets/4574058/946d432e-e2d8-4747-8b01-da307acd7f2f)

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

Medium

### What product/project are you using Modus Components for ?

e-Builder

### What is your team/division name ?

e-Builder

### Are you willing to contribute ?

Yes

### Are you using Modus Web Components in production ?

yes

apaddock:
https://app.clickup.com/t/868690163 

![image](https://github.com/trimble-oss/modus-web-components/assets/4574058/967955ff-e47e-483c-9f03-942cc9e67729)

rthanga1:
I hope this is done now. Good to close @yohernandez ?

yohernandez:
> I hope this is done now. Good to close @yohernandez ?

Yes, it can be closed.

PR #1948, opened 2023-12-12
Add density property to Modus Table

yohernandez:
## Description

- Add density property to Modus Table component.
- Update SCSS files to set the row height based on the property value: 
    - Relaxed (default) -> 3rem
    - Comfortable -> 2rem
    - Compact -> 1.5rem
- Add unit and e2e tests.
- Update Storybook

References #1930

## Type of change

- [X] Documentation update

## How Has This Been Tested?

Added a unit test and a e2e test to verify that a change in the density property update the table class properly. All tests passed sucessfully. 
Manually tested different scenarios in Storybook like badge, column resize and reorder, inline editing, row actions, expandable rows, row selection, etc.

## Checklist

- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works


stencil-workspace/src/components/modus-table/modus-table.scss
```diff
@@ -348,6 +348,33 @@ table {
...
+    }
+  }
+
+  &.density-compact {
+    tr {
+      height: 1.5rem;
+
+      td {
+        padding-top: 0.125rem;
+        padding-bottom: 0.125rem;
```
cjwinsor, line 370: when i ran `npm run lint` it corrected this and ordered them alphabetically

stencil-workspace/src/components/modus-table/modus-table.scss
```diff
@@ -348,6 +348,33 @@ table {
     right: 0;
     z-index: 1;
   }
+
+  &.density-relaxed {
+    tr {
+      height: 3rem;
```
cjwinsor, line 354: We should probably be using the helper variables here instead of hardcoding
```scss
$rem-2px: 0.125rem;
$rem-24px: 1.5rem;
$rem-32px: 2rem;
$rem-48px: 3rem;
```

cjwinsor: this applies to each place these are referenced.  I don't think we have to go so far to make this a variable in the global file at this time

stencil-workspace/src/components/modus-table/modus-table.scss
cjwinsor, line 1: Taking a look at these with checkboxes... it looks like if you applied compact's padding around the checkbox in comfortable then it would better align with the row height... which suggests you could reduce some for compacts as well.  Not sure the modus decision on this.  

The header height for compact and comfortable are the same, while relaxed is bigger.

The header height with sorting is one pixel more than without, but probably an issue prior to your change so not as big of an issue.

image

cjwinsor: This was my test script
```html
    <h3>Compact</h3>
    <div style="display:flex; gap:1rem">
      <modus-table density="compact" sort="true" full-width="true"></modus-table>
      <modus-table density="compact" sort="false" full-width="true"></modus-table>
      <modus-table density="compact" sort="false" full-width="true" row-selection="true"></modus-table>
    </div>
    <h3>Comfortable</h3>
    <div style="display:flex; gap:1rem; margin-top:1rem;">
      <modus-table density="comfortable" sort="true" full-width="true"></modus-table>
      <modus-table density="comfortable" sort="false" full-width="true"></modus-table>
      <modus-table density="comfortable" sort="false" full-width="true" row-selection="true"></modus-table>
    </div>
    <h3>Relaxed</h3>
    <div style="display:flex; gap:1rem; margin-top:1rem;">
      <modus-table density="relaxed" sort="true" full-width="true"></modus-table>
      <modus-table density="relaxed" sort="false" full-width="true"></modus-table>
      <modus-table density="relaxed" sort="false" full-width="true" row-selection="true"></modus-table>
    </div>
      <script>
        const columns = [
          {
            header: 'First Name',
            accessorKey: 'firstName',
            id: 'first-name',
            dataType: 'text',
            size: 150,
            minSize: 80,
            footer: 'Total',
          },
        ];
        const data = [
          {
            firstName: 'Fred',
          },
          {
            firstName: 'Bugs',
          },
          {
            firstName: 'Daffy',
          },
          {
            firstName: 'Tom',
          },
          {
            firstName: 'Daffy',
          },
        ];
        
        document.querySelectorAll('modus-table').forEach((table) => {
          table.columns = columns;
          table.data = data;
        });
    </script>
```

cjwinsor (changes requested):

stencil-workspace/src/components/modus-table/modus-table.scss
yohernandez: Added the compact's padding in comfortable.
The table header must be the same for all density options. Updated the CSS to override the height for body rows only. 

image

The table row heights should be the same with and without the checkbox column for relaxed and comfortable. The case of compact is different because the table component uses the regular size version of the checkbox which is $rem-24px. UX wants a 2px padding top and bottom.

yohernandez: Also, we plan to allow users to select the checkbox size. This will allow us to use the small version for compact.

cjwinsor (approved):

Issue #1269, opened 2023-05-10
Modus NavBar - Slot in user menu for custom content


cjwinsor:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have read the [contributing guidelines](https://github.com/trimble-oss/modus-web-components/blob/main/CONTRIBUTING.md)

### Proposal

We were given the following UX, shown below. We will ensure alignment with Modus Design System and Modus Web Components teams before starting work

The scope of this change would be to add a `slot` under the profile menu links.  We will use that for user delegation within e-Builder, but would leave this area generic to allow it to be flexible.

![image](https://github.com/trimble-oss/modus-web-components/assets/1130129/18fbb1c5-2bf7-49a2-8c44-6e1e4bd87875)

**Tasks**
1. Add `slot` area to User Profile Menu
2. Handle styling when there is and isn't content to be added.

It is e-Builder's intent to work on this request.

### Motivation and context

We have a user delegation feature in e-Builder which allows a user to work on behalf of another individual. In order to maintain that feature and use the Modus NavBar, we require an area enable user context switching.  By making it a slot, we enable reusability and flexibility with its use.

### Which npm package?

@trimble-oss/modus-web-components

### Priority

Medium

### What is the product name the feature is requested for?

e-Builder

### What is the team name the product belongs to?

e-Builder

### What is the division name the team belongs to?

e-Builder

msankaran0712:
This is to reaffirm that Modus has approved the functionality for development.

cc @cjwinsor 

Note: Modus team tag the approved issues with `modus-approved`

msankaran0712:
Moving this to the backlog as the issue not been worked up on currently

msankaran0712:
@cjwinsor any updates on when are we taking this issue?

prashanth-offcl:
Closing this issue as the changes are already merged

PR #1954, opened 2023-12-13
Add Slot in user profile menu for custom content

Sanket8793:
## Description

Add slot area to User Profile Menu

References #1269 

## Type of change

- [X] Documentation update

## How Has This Been Tested?

we have to enable user context and then render the html you want to show using slot userCustomContext

## Checklist

- [X] Any dependent changes have been merged and published in downstream modules


stencil-workspace/src/components/modus-navbar/modus-navbar.tsx
```diff
@@ -476,7 +480,11 @@ export class ModusNavbar {
                       reverse={this.reverse}
                       username={this.profileMenuOptions?.username}
                       variant={this.variant}
-                    />
+                    >
```
cjwinsor, line 483: When i ran `npm run lint` it corrected this to 
```
                      variant={this.variant}>
                      {this.enableUserContent && <slot name={this.SLOT_USERCUSTOMCONTENT}></slot>}
```

stencil-workspace/src/components/modus-navbar/modus-navbar.e2e.ts
```diff
@@ -400,4 +400,40 @@ describe('modus-navbar', () => {
     expect(await primaryLogo).toBeFalsy();
     expect(await secondaryLogo).toBeFalsy();
   });
+
```
cjwinsor, line 403: When i ran `npm run lint` it found 4 issues in the test file and fixed them.  Lines 403, 420, 422, and 438 all have an extra space character.

stencil-workspace/storybook/stories/components/modus-navbar/modus-navbar.stories.tsx
```diff
@@ -67,6 +76,7 @@ const Template = ({ profileMenuOptions, buttons, showSearch, enableSearchOverlay
       </modus-list>
 
     <div slot="notificationMenu">Render your own notification menu.</div>
+    <div slot="userCustomContent">Render your own custom content.</div>
```
cjwinsor, line 79: This doesn't actually show up in the example, and i think in this case it should.

stencil-workspace/src/components/modus-navbar/modus-navbar.tsx
```diff
@@ -49,6 +49,9 @@ export class ModusNavbar {
   /** (optional) Whether to show search overlay or not. */
   @Prop() enableSearchOverlay: boolean;
 
+  /** (optional) Whether to show user content in the profile menu or not. */
+  @Prop() enableUserContent: boolean;
```
cjwinsor, line 53: Do we need a property for this? It seems like if they don't provide the slot, it doesn't produce any side affects.  If we need to have the property for some reason, lets discuss the property name.

cjwinsor (changes requested):

stencil-workspace/src/components/modus-navbar/modus-navbar.tsx
Sanket8793: @cjwinsor I've used this property as a flag to hide and show the slot in the profile menu, but I'll remove this if it's not necessary. Want your opinion on this

cjwinsor: @Sanket8793 I think we can remove the property. It will function the same as the slot for notifications.  We can always add it in the future if there is a strong need, harder to remove it when there wasn't any need. The extra elements aren't that costly and we benefit from a simpler implementation and usage.

stencil-workspace/src/components/modus-navbar/modus-navbar.tsx
```diff
@@ -140,6 +143,7 @@ export class ModusNavbar {
 
   readonly SLOT_MAIN = 'main';
   readonly SLOT_NOTIFICATIONS = 'notifications';
+  readonly SLOT_USERCUSTOMCONTENT = 'userCustomContent';
```
cjwinsor, line 146: can we instead name this `SLOT_PROFILE_MENU` and have the value be `profileMenu`

Sanket8793: Yes I updated the name

cjwinsor (approved):

Issue #1963, opened 2023-12-19
Add GroupRight slot to Modus Content Tree items


yohernandez:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have read the [Modus style guidelines](https://modus.trimble.com/components/web/introduction/)

### Proposal

Currently, the modus content tree component offers only a single slot for the label of the content tree items. 
This epic is to update the content tree to offer a slots - “groupRight” slot (similar to the slots for the modus table) 

![image](https://github.com/trimble-oss/modus-web-components/assets/48927186/b5cc7b1c-0b3c-4588-b5e8-38031f8e7e17)

This would allow the label slot to be used when only a label is needed but support actions on content where needed. 

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

Medium

### What product/project are you using Modus Components for ?

e-builder

### What is your team/division name ?

e-builder

### Are you willing to contribute ?

Yes

rthanga1:
@cjwinsor To understand the requirement and the need

cjwinsor:
@Roliver-Javier @yohernandez Before I dive into this review I have a few questions:

 - What is the purpose of the `groupLeft` slot. What is the use case for its addition?
 - What is the intended goal of `groupRight` slot. What is the use case for this addition?
 
 I want to make sure we aren't over complicating without cause.

rjavier-trimbler:
Hi @cjwinsor 

the purpose for each group is to have content in both sides, currently we only have room to place a label and that slot has an specific role for this. The group-left slot would be for components with its own role in order to not generate confusion with the screen reader if we use the label slot as a content on the left

apaddock:
<img width="781" alt="image" src="https://github.com/trimble-oss/modus-web-components/assets/4574058/a9c799c6-b44b-47d8-9f44-ca0026cad899">

PR #1986, opened 2023-12-29
Modus Content Tree: adding  right slot  to modus tree view item

rjavier-trimbler:
## Description
- adding  group-right slot  
- adding  group-left slot 

the purpose for each group is to have content in both sides, currently we only have room to place a label and that slot has an specific role for this. The group-left slot would be for components with its own role in order to not generate confusion with the screen reader if we use the label slot as a content on the left.

There's a label slot to the left already but for accessibility purposes if I use that slot and add another component
for example:
a button. it will place a Role as a label instead of a role as a button and if I place a role in the button component it will tab navigate twice (first the label slot container and then the Button component with the other role).

and that's why we have included the group-left where the responsability to place a correct role for tab navigation will be in the component that we're adding not the slot container

References #1963 

## Type of change

- [x] Documentation update

## Checklist


cjwinsor:
@Roliver-Javier @yohernandez I'm going to add comments to the the Issue so we can discuss the goal of this there.

apaddock:
@cjwinsor what's the status of the review for this PR?

cjwinsor:
> @cjwinsor what's the status of the review for this PR?

I have been pulled off this PR multiple times due to other requests to help with other PRs from e-Builder, but I was able to spend some time late last night to formulate a plan.  I will be adding additional comments this afternoon.  We will not be moving forward with the left slot unless justified as a need.  The right slot will need to be replaced with an actual action menu component which I have started the work on in order to assist e-Builder in a more timely resolution to this PR.

cjwinsor:
2 main points
1. We shouldn't be adding a left slot. Please provide the exact usage scenario that we can address in another issue if valid.
2. Instead of a slot on the right side, it needs to be a proper component.  The guidelines for this component do mention action buttons on the right side, so this is good to become an actual prop on the item.

The end result should look as shown in the screenshot on the issue
image

To achieve this, I recommend creating a `ModusActionBar` component that takes a property for number of visible items, size (small, medium, large), the array of actions (id, icon, label) and emits an event when an action is clicked.  

This component should not go into storybook at this time, though we may decide to elevate its status in the future.

Utilize popperjs to position the modus-list in the overflow menu.

Message me directly when you are ready to work on this as I have created an action bar component which could serve as a basis to this work.  Its very POC, but illustrates everything mentioned above.

This is a screenshot of my current POC (borders added around the `modus-action-bar` for clarity, and updated list items to have a `leftIcon` property so the list item has an icon... just a part of lists that hasn't been added yet but is in the guidelines.)
image

There are a few UX considerations that would ultimately need addressed but I will approve this PR once we can define actions per `modus-tree-view-item` which get passed into the `modus-action-bar` which handles creating the action bar.

The `modus-tree-view-item` would hard code the number of visible items to 3 (3 actions or 2 actions + overflow menu) so that only 3 icons total can appear.

cjwinsor:
As discussed with @Roliver-Javier this PR will be closed. A new PR will be opened to address the comments from this PR as the changes were significant.

cjwinsor:
As discussed with @Roliver-Javier this PR will be closed. A new PR will be opened to address the comments from this PR as the changes were significant.

Issue #1982, opened 2023-12-27
Autocomplete: missing Watch decorator for property 'value'


yohernandez:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/trimble-oss/modus-web-components/blob/main/CONTRIBUTING.md)

### Describe the issue

Changing the value of the autocomplete property "value" using javascript is not being honored. The options are filtered by the previous value. The property "value" has no @Watch decorator. 

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

Medium

### What product/project are you using Modus Components for ?

e-builder

### What is your team/division name ?

e-builder

### Are you willing to contribute ?

Yes

rthanga1:
Closing this one as the PR is merged.

PR #1993, opened 2024-01-03
Modus Autocomplete: Adding @watch decorator to value property

rjavier-trimbler:
## Description

Issue Fixed:
- when we change the value property using javascript with empty string after selecting an option it was showing the option selected in the list instead of the list options.

- when we change the value property using javascript with a value that not match with any of the list, it was showing the first element of the list instead of the "Not Found message"

Both scenarios were fixed for Default and Custom options

References #1982 

## Type of change

- [ ] Documentation update

## Checklist


rthanga1 (dismissed):

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
```diff
@@ -87,7 +87,18 @@ export class ModusAutocomplete {
 
   /** The autocomplete's search value. */
   @Prop() value: string;
```
cjwinsor, line 89: @Prop are immutable by default (It is convention, still works to change it but it throws a warning). In order to let stenciljs know that this prop can be changed both externally and internally, we need to set the mutable option on the prop.
```js
@Prop({ mutable: true }) value: string;
```

If you want to go the extra mile, we need to add this to the `modus-text-input`'s `value` as well. That will eliminate a lot of warning messages.

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
```diff
@@ -87,7 +87,18 @@ export class ModusAutocomplete {
...
+  @Watch('value')
+  watchValue(newValue: string) {
+    if (newValue === '') {
```
cjwinsor, line 92: Should this be something like `this.value !== newValue`?

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
```diff
@@ -137,16 +149,37 @@ export class ModusAutocomplete {
...
-  displayNoResults = () =>
-    this.showNoResultsFoundMessage &&
-    this.hasFocus &&
-    !this.visibleOptions?.length &&
-    !this.visibleCustomOptions?.length &&
-    this.value?.length > 0;
+  handleNoResults = (value: string) => {
+    let handleResults;
+    if (this.options?.length > 0) {
+      handleResults =
```
cjwinsor, line 155: If we don't care about the results, we can use `some` instead of `filter`
```js
    handleResults =
        (this.options as ModusAutocompleteOption[]).some((o: ModusAutocompleteOption) => {
          return o.value.toLowerCase().includes(value.toLowerCase());
        }).length > 0;
```

Same with `this.customOptions`.

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
```diff
@@ -137,16 +149,37 @@ export class ModusAutocomplete {
...
+
+  displayNoResults = () => {
+    if (this.showNoResultsFoundMessage && this.hasFocus && !this.visibleOptions?.length && this.value?.length > 0) {
+      if (this.containsSlottedElements) {
+        return this.hasNoResults;
+      }
+      return true;
+    }
+
+    return this.hasNoResults && this.visibleOptions?.length > 0 && this.hasFocus;
```
cjwinsor, line 177: When would `this.visibleOptions.length > 0` when `this.hasNoResults` is true?

cjwinsor: And is there any redundancy now that you added `handleNoResults`?  Did `displayNoResults` not work to show the panel?

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
```diff
@@ -137,16 +149,37 @@ export class ModusAutocomplete {
...
+      handleResults =
+        this.customOptions.filter((o: any) => {
+          return o.getAttribute(DATA_SEARCH_VALUE).toLowerCase().includes(value.toLowerCase());
+        }).length > 0;
+    }
+    this.hasNoResults = !handleResults;
+  };
+
+  displayNoResults = () => {
+    if (this.showNoResultsFoundMessage && this.hasFocus && !this.visibleOptions?.length && this.value?.length > 0) {
```
cjwinsor, line 170: Before it was checking both `visibleOptions` and `visibleCustomOptions` but now you're only checking `visibleOptions`

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
```diff
@@ -254,7 +287,11 @@ export class ModusAutocomplete {
         aria-readonly={this.readOnly}
         aria-required={this.required}
         class={classes}
-        onFocusin={() => (this.hasFocus = true)}>
+        onFocusin={() => {
+          this.hasFocus = true;
+          this.updateVisibleOptions(this.value);
```
cjwinsor, line 292: Are these updates here on focus needed.  I'm concerned we are triggering a lot of rendering and updating.

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
cjwinsor: You might be able to actually structure this without the @Watch.  I'm looking into it.

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
rjavier-trimbler: inside that function is where we play with the visibility of the options and in this case right here, if there's no value in the textInput when you're focusin it will show all the options again.

because this line:
```
  if (!search || search.length === 0) {
      this.visibleOptions = this.options as ModusAutocompleteOption[];
      return;
    }
```

cjwinsor (approved):

Issue #1995, opened 2024-01-04
Modus table: maxSize is ignored when full-width is true


janwiebe-glasbergen:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/trimble-oss/modus-web-components/blob/main/CONTRIBUTING.md)

### Describe the issue

When using the table component in the full-width mode, the max-width property is ignored.

This is currently causing unwanted behavior when having rows with longer content.

**Steps to reproduce:**
* Create a table component
* Set full-width property to true
* Configure the maxSize property on the column
* Enter a value that causes a cell bigger than the max width.

![image](https://github.com/trimble-oss/modus-web-components/assets/116563683/4fba4103-6655-4d36-9f7f-fd74f5e2c9b4)

**Expected behavior**
maxSize limit's the amount of text shown in the cell, even when full-width is set to true

### What is the issue regarding ?

@trimble-oss/modus-web-components

### What version of npm package are you using ?

0.10.0

### Priority

Medium

### What product/project are you using Modus Components for ?

Nova / Construction calculation platform UI

### What is your team/division name ?

ACDC

### Are you willing to contribute ?

None

PR #2001, opened 2024-01-09
Table: Set table-layout to fixed for full width table

janwiebe-glasbergen:
## Description

Changed table-layout to fixed, as table-layout: auto ignores the column (max-)width.

Before:
image

After:
image

References #1995

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

Tested by running in Stencil:
- Create a table
- Set fullWidth: true
- Create a column with a max-width
- Add a value that is longer than the max-width
- Check if result value is concatenated.

## Checklist
- [x] I have made corresponding changes to the documentation _(no docs updated needed)_


rthanga1 (dismissed):
Verified with the following piece of code

[sample-table.txt](https://github.com/trimble-oss/modus-web-components/files/14038941/sample-table.txt)

cjwinsor:
@janwiebe-glasbergen I finished my review of this and want to take it back to the team real quick before approving and merging.  If nothing comes from the team tomorrow we can proceed.  I'll comment again tomorrow by noon EST.

Due to how the default column width is `150px`, in this scenario where the table width is 100%, the resulting real column widths will be based on their ratio of defined sizes.  So a table where you define the first column with a size of `150px` and leave the second column undefined you will get 2 columns that take of 50% of the width each.  The first column won't be `150px` and the second won't be taking the rest of the space.  In order for this to work, the `size` of at least one column would need to be `unset`.  I'm not sure what downstream affects that might cause with other features though.  Resize columns in the `full-width` scenario already is an interesting experience.

cjwinsor (approved):

Issue #2035, opened 2024-01-16
Add Active/Pressed State to Navbar Buttons


Sanket8793:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/trimble-oss/modus-web-components/blob/main/CONTRIBUTING.md)

### Describe the issue

Currently all navbar buttons have no add active/pressed state so they need to be added on button click.

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

Medium

### What product/project are you using Modus Components for ?

E-Builder/Branding

### What is your team/division name ?

E-Builder/DarkKnights

### Are you willing to contribute ?

Yes

cjwinsor:
This ticket was discussed in a meeting with the two teams.  I'm waiting to hear if they need technical help to resolve their need here after they check their options with gainsight.

PR #2036, opened 2024-01-16
Navbar: Add Active/Pressed State to buttons

Sanket8793:
Add Active/Pressed State to Navbar Buttons

References #2035 

## Type of change

- [ ] Documentation update

## How Has This Been Tested?
  
Make sure that when button is pressed, it should display an active state

## Checklist


cjwinsor:
@Sanket8793 The help doesn't have a menu, so with this change it should just show it as active until someone clicks somewhere else.  Is there a use case for this.  The issue seems to be for the navbar buttons, but it was too generic and lacks screenshots or details.  Does this have to do with e-Builders resource center and needing it to mark the help icon as active while that is open?  If so, we might need a solution that doesn't set the icon as pressed under every state.

DafneeZambrano:
Hi Chris, for buttons that don't open a menu (like search) we use a quick press. So the state is active while the user is pressing the button. This is an accessibility requirement to indicate the action to the user. 

If the button triggers a menu, then it stays in an active state until the user closes the menu.

For the help menu, we will eventually trigger the Gainsight menu, so we will need it as an active state then.

rthanga1 (changes requested):
Holding for Chris's comment get addressed

cjwinsor:
> Hi Chris, for buttons that don't open a menu (like search) we use a quick press. So the state is active while the user is pressing the button. This is an accessibility requirement to indicate the action to the user.
> 
> If the button triggers a menu, then it stays in an active state until the user closes the menu.
> 
> For the help menu, we will eventually trigger the Gainsight menu, so we will need it as an active state then.

@DafneeZambrano The current implementation keeps the button active even after the click action is completed.  So it remains marked as active until another element is interacted with.  So if a popup were to be triggered then it remains active until the user comes back to the page and interacts with anything else, which is a strange behavior.  

I agree that while its being clicked it should be active, and if it is showing something then ideally its still marked active as well, though I need a moment to consider how best to work with your desire to have a gainsight menu show at that location and keep the icon active at the same time. I will come back to this PR later in the day.

cjwinsor:
There is a meeting to discuss the particulars on 1/24 at 11am EST.

cjwinsor:
This ticket was discussed in a meeting with the two teams. I'm waiting to hear if they need technical help to resolve their need here after they check their options with gainsight.

cjwinsor:
@Sanket8793 I see you are adding a bunch of icon files, perhaps you had a bad merge of main?  Can you fix your branch?  Also I don't seem to see any different handling for the help icon not being marked active after click, was there any intended change there?

zabdur-rahman:
> @Sanket8793 I see you are adding a bunch of icon files, perhaps you had a bad merge of main? Can you fix your branch? Also I don't seem to see any different handling for the help icon not being marked active after click, was there any intended change there?

@cjwinsor We won't be adding special handling for the help icon. An issue will be made with Gainsight to address it.

cjwinsor:
@zabdur-rahman @Sanket8793 PR still has issue with having all those extra files added for icons. Please sort out your branch so it only has your changes and doesn't have the special handing for help, then we can move forward.

cjwinsor:
@Sanket8793 The latest Modus Web Components switched out the `IconMap` for the `ModusIconMap`, which gives all the Modus Icons a pressed state.  The Navbar now uses `ModusIconMap`.  Is this PR necessary?  I'm actually looking to see if we can move away from using the icons in `svgs` in favor for the full icon list available from `ModusIconMap`  and the `generated-icons`.

Sanket8793:
@cjwinsor Since the Navbar uses modus icons from ModusIconMap we don't need these changes, I'm going to close this PR

cjwinsor (approved):

Issue #1940, opened 2023-12-07
Modus Table - Make the header row sticky at the page level 


msankaran0712:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have read the [Modus style guidelines](https://modus.trimble.com/components/web/introduction/)

### Proposal

Users can currently enable a vertical scroll on the table by adjusting the `maxHeight` parameter. When scrolling through the table, the header row remains fixed, as shown in the screenshot (https://modus-web-components.trimble.com/?path=/story/components-table--pagination&args=maxHeight:700px). However, the desired behavior is for the header to remain "sticky" even when scrolling through the entire page, as shown in the example (https://modus-bootstrap.trimble.com/examples/table-with-sticky-header-and-column/).

Current behaviour screenshot:
![image](https://github.com/trimble-oss/modus-web-components/assets/50131391/a428dba6-d7f7-499c-b415-35633bf2a64d)

Expected behaviour:
![image](https://github.com/trimble-oss/modus-web-components/assets/50131391/b8d80942-da78-48b2-8a38-ed2c3fc24678)

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

Medium

### What product/project are you using Modus Components for ?

E-Builder

### What is your team/division name ?

E-Builder

### Are you willing to contribute ?

Yes

rthanga1:
https://modus-bootstrap.trimble.com/examples/table-with-sticky-header-and-column/ -> This behaviour should be the default.
When `sticky-top` flag is set, then the table should have the sticky header.

apaddock:
@cjwinsor we're going to begin this story now, is there any feedback on implementation? 
If I'm understanding the ask, we want a way to keep headers fixed when scrolling the page regardless of the max-height setting.

cjwinsor:
@apaddock the `position: sticky` is the modern approach to this. It will stick to the top of whatever its scroll container is.

enowak1031:
Deprioritizing. This is an old issue. We have sticky headers currently. Unless someone brings up the issue again, we are not going to work on it.

PR #2039, opened 2024-01-17
Modus Table: table header was not sticky when scrolling through the entire page

rjavier-trimbler:
## Description
When an element has overflow: auto, it can act as a block container and affect the positioning relationship of child elements, including those with position: sticky. that's why that after removing this line in the .table-container class, it let the child element (the `<th>` in this case)  to be sticky.

References #1940 

## Type of change

- [ ] Documentation update

## Checklist


rthanga1:
@cjwinsor Do we have to push the issue to backlog or is someone going to try another approach?

cjwinsor:
> @cjwinsor Do we have to push the issue to backlog or is someone going to try another approach?

I was going to bring up in the next standup. We are still communicating that this solution didn't pan out through e-builder so I'll see what they are thinking.

Issue #1998, opened 2024-01-04
Table: Add support for wrapping text


cjwinsor:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have read the [Modus style guidelines](https://modus.trimble.com/components/web/introduction/)

### Proposal

The table should allow for the text to wrap, instead of being truncated.  Since the current table is setup to truncate by default, we should leave that.  My suggestion is to add a `wrap-text` property that defaults to `false`.  When enabled, it will remove the css that currently causes the body text to be truncated with ellipsis.  I do not expect this setting to affect the header or the footer.

Additionally, padding needs to be maintained.  Currently, the `<tr>` has its height set relative to its `density`.  Each column type should ensure that it's effective padding is maintained between wrapping and non-wrapping configurations.

Currently looks like this:
![image](https://github.com/trimble-oss/modus-web-components/assets/1130129/e82e1d1b-c11b-484c-9647-9bc3d435e6c6)

After changes, it could look like this:
![image](https://github.com/trimble-oss/modus-web-components/assets/1130129/b7dd3400-ce8b-4b3d-a405-3d8c51eec5fd)

### Motivation and context

Currently columns are displayed which can show a large block of text.  Since this text is truncated, users cannot scan or otherwise search for a specific row based on the value of the cell.  By adding back text wrapping, the table can function the same as a standard table and gives us the control on whether its more valuable to have fixed and predictable row heights, or all the text in the cells being displayed with irregular row heights.

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

Medium

### What product/project are you using Modus Components for ?

e-Builder

### What is your team/division name ?

e-Builder

### Are you willing to contribute ?

None

apaddock:
Approved design submission to modus: https://modus.ideas.aha.io/ideas/MOD-I-19

rthanga1:
Ewa has mentioned in the Aha that we expose a property to support wrap line to the table.
But I think the property should go to the column rather than the table itself. 
I am thinking exporting these properties to each columns

1. noWrap - Default false, when set to true the text goes to next line when the size goes beyond the width of the cell
2. maxHeight  - by default it will be set to 0. When a value is set (say for ex 50px), then the row can expand upto the set height

apaddock:
[...]

PR #2056, opened 2024-01-23
Table: Add support for wrapping text

bradley-eb:
## Description
The table should allow for the text to wrap, instead of being truncated. Since the current table is setup to truncate by default, we should leave that. My suggestion is to add a wrap-text property that defaults to false. When enabled, it will remove the css that currently causes the body text to be truncated with ellipsis. I do not expect this setting to affect the header or the footer.

References #1998

## Type of change

- [ ] Documentation update

## Checklist


dnarey:
Feedback from UX:
Here are the expectations
1. The checkboxes should be top aligned
2. The Text/anything in all the cells should be top aligned
3. Hyperlinks also have to be truncated.

apaddock:
@bradley-eb fix the following issues (per ux) 

- [ ]  The checkboxes should be top aligned

- [ ] The Text/anything in all the cells should be top aligned

- [ ] Hyperlinks also have to be truncated.

JTrimbleR:
Validated test scenarios after feedback, working as expected.

Pre-conditions: wrapText control set to true, columnResize control set to true

Scenario 1: Resize a column which text’s length exceeds the min width to min width, text should wrap (no truncation) and all other field in the row also resize

Scenario 2: Resize a column which text’s length exceeds the max width to max width, text should wrap (no truncation) and all other field in the row also resize

Scenario 3: Validate that columns that have text with special characters are able to wrap (&, %, “, ')

Scenario 4: Validate that a column whose text is wrapping does not affect the size of other columns

Scenario 5: Checkboxes

Scenario 6: Badges

Scenario 7: Column Resize

Scenario 8: Hyperlinks

Scenario 9: Borderless table

Scenario 10: Value Formatter

Scenario 11: Pagination

Scenario 12: Expandable rows

Scenario 13: Densities

stencil-workspace/src/components/modus-table/parts/cell/modus-table-cell.tsx
```diff
@@ -23,6 +23,7 @@ export const ModusTableCell: FunctionalComponent<ModusTableCellProps> = ({ cell,
       class={`
       ${hasRowsExpandable || frozenColumns.includes(cell.column.id) ? 'sticky-left' : ''}
       ${cell.column.getIsResizing() ? 'active-resize' : ''}
+      ${wrapText ? 'align-top' : ''}
```
cjwinsor, line 26: We shouldn't have to add a css class to every `td`.  Ideally no extra class is needed to do the padding with or without text-wrap.  Given the density class and wrapped text classes though, you should be able to manage without the extra class.

Below is something just to give an idea of how it can be done.
```scss
table {
  &.density-relaxed {
    &.truncate-text tbody tr td {
      //...
    }
    
    &.wrap-text tbody tr td {
     //...
    }
  }
}
```

cjwinsor (changes requested):
Make sure this pr only has your changes (i see a change to the readme of modus-icons).  I would like to avoid adding classes to every cell of tbody.

stencil-workspace/src/components/modus-table/parts/cell/modus-table-cell.tsx
bradley-eb: The classname and property `align-top` has been removed.

bradley-eb:
> Make sure this pr only has your changes (i see a change to the readme of modus-icons). I would like to avoid adding classes to every cell of tbody.

@cjwinsor Thanks for your review. I have removed unnecessary changes, fixed all lint issues as well, and got rid of `align-top`.

JTrimbleR:
Validated scenarios against latest changes, working as expected.

Pre-conditions: wrapText control set to true, columnResize control set to true

Scenario 1: Resize a column which text’s length exceeds the min width to min width, text should wrap (no truncation) and all other field in the row also resize

Scenario 2: Resize a column which text’s length exceeds the max width to max width, text should wrap (no truncation) and all other field in the row also resize

Scenario 3: Validate that columns that have text with special characters are able to wrap (&, %, “, ')

Scenario 4: Validate that a column whose text is wrapping does not affect the size of other columns

Scenario 5: Checkboxes

Scenario 6: Badges

Scenario 7: Column Resize

Scenario 8: Hyperlinks

Scenario 9: Borderless table

Scenario 10: Value Formatter

Scenario 11: Pagination

Scenario 12: Expandable rows

Scenario 13: Densities

stencil-workspace/src/components/modus-table/modus-table.scss
```diff
@@ -197,16 +197,31 @@ table {
     tr {
       color: $modus-table-body-color;
 
-      .wrap-text {
+      .truncate-text {
         overflow: hidden;
         text-overflow: ellipsis;
         text-wrap: nowrap;
```
cjwinsor, line 203: Change `text-wrap: nowrap` to `white-space: nowrap`, this will be important for the test update suggestion I am going to make as `text-wrap` isn't a fully supported property so my test suggestion won't work with it.  Also the rest of MWC uses `white-space: nowrap`.

stencil-workspace/src/components/modus-table/modus-table.e2e.ts
```diff
@@ -935,4 +935,28 @@ describe('modus-table', () => {
...
+      ...MockData,
+    ]);
+    component.setProperty('wrapText', true);
+
+    await page.waitForChanges();
+
+    const tableData = await page.findAll('modus-table >>> td');
+    const firstItem = tableData[0].textContent;
+
+    expect(firstItem).toBe('This is an example of long text');
```
cjwinsor, line 960: This will always assert true as the content isn't affected by the truncation. This is only visual.

My suggestion is to update the test so it validates that `wrap-text` is applied and that the styles are correct.  We also need another test that validates when `wrapText` is false.

Write a selector that finds the span, so you can check `wrapText = false`
```js
    const style = await spanData.getComputedStyle();
    expect(style['overflow']).toBe('hidden');
    expect(style['text-overflow']).toBe('ellipsis');
    expect(style['white-space']).toBe('nowrap');
```

and for `wrapText = true`

```js
   const style = await spanData.getComputedStyle();
    expect(style['overflow-wrap']).toBe('break-word');
    expect(style['word-break']).toBe('break-word');
```

cjwinsor (changes requested):

stencil-workspace/src/components/modus-table/modus-table.scss
bradley-eb: Updated.

Issue #2066, opened 2024-01-25
Modus-autocomplete: Hitting Space should also select a focused option as well as hitting Enter


yevhenkravets:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have read the [Modus style guidelines](https://modus.trimble.com/components/web/introduction/)

### Proposal

- Space/ Enter Key: Selects the option from the dropdown and adds the text to the textbox.

This is a part of the approved Modus submission: https://modus.ideas.aha.io/ideas/MOD-I-14

### Motivation and context

This is partially related to #979 

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

Medium

### What product/project are you using Modus Components for ?

E-builder

### What is your team/division name ?

E-builder

### Are you willing to contribute ?

Yes

PR #2067, opened 2024-01-25
Autocomplete: Hitting Space should also select a focused option as well as hitting Enter

yevhenkravets:
## Description

Space/ Enter Key: Selects the option from the dropdown and adds the text to the textbox.
This is a part of the approved Modus submission: https://modus.ideas.aha.io/ideas/MOD-I-14

Also added accessibility attributes following these specifications:
https://w3c.github.io/aria/#combobox
https://w3c.github.io/aria/#aria-autocomplete
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role

References #2066

EPD-11997

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

Added relevant integration tests and it will be tested also by the e-builder QA through the storybook.

## Checklist


stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
```diff
@@ -264,6 +266,10 @@ export class ModusAutocomplete {
       type="search"
       value={this.value}
       onBlur={this.handleInputBlur}
+      role="combobox"
+      aria-autocomplete="list"
+      aria-controls={this.listId}
+      aria-expanded={this.displayOptions()}
```
yevhenkravets, line 272: Added accessibility attributes following specifications I added in the description.

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
```diff
@@ -170,7 +171,8 @@ export class ModusAutocomplete {
 
   handleOptionKeyPress = (event: any, option: any, isCustomOption = false) => {
     this.disableFiltering = !this.disableCloseOnSelect;
-    if (event.key !== 'Enter') {
+    const allowedKeys = ['Enter', ' ']; // Hitting Enter or Space should select the option
+    if (!allowedKeys.includes(event.key)) {
```
yevhenkravets, line 175: Hitting Space also selects an option now.

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
```diff
@@ -111,6 +111,7 @@ export class ModusAutocomplete {
   @State() customOptions: Array<any> = [];
   @State() visibleCustomOptions: Array<any> = [];
   @State() disableFiltering = false;
+  private listId = `autocomplete-${Math.random().toString().slice(2, 7)}`;
```
rthanga1, line 114: Can you explain what this does?

rthanga1:
@yevhenkravets Can you explain what this PR does? 
1. I opened up the auto complete component in the storybook
2. Used tab to focus the component
3. And tried using 'Space' and 'Enter' key which doesn't seem to do anything

My understanding is, when the 'Enter' key is pressed, it should open up the options, right?

yevhenkravets:
> @yevhenkravets Can you explain what this PR does?
> 
> 1. I opened up the auto complete component in the storybook
> 2. Used tab to focus the component
> 3. And tried using 'Space' and 'Enter' key which doesn't seem to do anything
> 
> My understanding is, when the 'Enter' key is pressed, it should open up the options, right?

No, the expectation here is when you go through the list of options and press Enter/Space then the currently focused option gets selected.

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
yevhenkravets: It generates a dynamic ID for the list element. It's not static to be able to use multiple autocomplete components on the page and every list element has a unique ID.
But, the main reason why we need the ID is for accessibility specifications. We use it in aria-controls attribute here https://github.com/trimble-oss/modus-web-components/pull/2067/files#diff-fd75ee2b4254f5561ecb52008b06050dc5ce305354677d36d07caed91e29be61R271
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls

cjwinsor: @yevhenkravets This doesn't work as it doesn't guarantee uniqueness.  

You'll need to create a utility function for generating IDs.  My suggestions are 

- create IDs with the prefix of `mwc_id_` following by the next value
- start at 0
- import utility function into component to create the ID
- test for utility function

This will have a lot of value as we start doing accessibility work.  

Take this generated id and append `_list` after it to further clarify. 

We do have a `createGuid()` utility as well.. but that will create very verbose IDs and ideally for our purposes a count is all that is required as we don't' have or need persistence.

cjwinsor (changes requested):

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
yevhenkravets: @cjwinsor Updated the code as you suggested.
Screenshot 2024-01-30 at 16 47 53

cjwinsor (approved):

Issue #2088, opened 2024-01-30
User Menu: Hover State has no blue border/outline


ausnewland11:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have read the [Modus style guidelines](https://modus.trimble.com/components/web/introduction/)

### Proposal

There is no blue outline & blue border on hover for the User menu and this should be added.

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

Medium

### What product/project are you using Modus Components for ?

Modus Navbar User Menu

### What is your team/division name ?

Dark Knights

### Are you willing to contribute ?

Yes

PR #2090, opened 2024-01-30
Navbar: Add hover border/outline to profile menu icon

ausnewland11:
## Description

Added outline and border color to the profile menu icon when in a hover state.

References [#2088](https://github.com/trimble-oss/modus-web-components/issues/2088) 

## Type of change

- [ ] Documentation update

## Checklist

### DIFF
--- a/stencil-workspace/src/components/modus-navbar/modus-navbar.scss
+++ b/stencil-workspace/src/components/modus-navbar/modus-navbar.scss
@@ -156,13 +156,14 @@ nav {
           width: 32px;
         }
 
-        > :first-child:hover,
-        > :first-child:focus {
-          border: $rem-4px solid $modus-navbar-icon-color;
+        > :first-child:hover {
+          border: $rem-4px solid $modus-navbar-icon-hover-bg;
         }
 
-        > :first-child:active {
-          outline: $rem-4px solid $modus-navbar-profile-icon-active-border-color;
+        > :first-child:active,
+        > :first-child:focus {
+          border: $rem-4px solid $modus-navbar-profile-icon-active-border-color;
+          outline: none;
         }
 
         &:hover {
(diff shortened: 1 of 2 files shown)


stencil-workspace/src/components/modus-navbar/modus-navbar.scss
```diff
@@ -228,6 +224,13 @@ nav {
         > :first-child:active {
           outline: $rem-4px solid $modus-navbar-blue-profile-icon-active-border-color;
         }
+
+        &:hover {
+          cursor: pointer;
```
cjwinsor, line 229: Properties should be in alphabetical order, running `npm run lint` in the `stencil-workspace` will resolve this and any other issues.

cjwinsor:
@ausnewland11 I'm not sure this matches what the UX expected. Can you work with your UX to verify this is as intended.

cjwinsor (changes requested):

ausnewland11:
> @ausnewland11 I'm not sure this matches what the UX expected. Can you work with your UX to verify this is as intended.

@cjwinsor I checked with UX and have updated the styling.

cjwinsor (commented):
I'm currently in the middle of reviewing.  I want to validate with UX that the Focus and Hover+Active states are correct.
image

What I will be checking next:

- Light and Dark themes (I know you didn't update, but we should be for consistency)
- With and without avatar image (I only tested with initials so far)

cjwinsor:
@ausnewland11 For the active+hover state, i'm expecting it not to have an extra thick border and to match active state.  Also can you apply similar changes to light and dark themes so they aren't left out and everything is consistent.

DafneeZambrano:
@cjwinsor @enowak1031 @egunther39 I brought this to the UX team. We are ok with the focus, hover, and active states. We do not need "Active + hover" as the active state does not need to change on hover. It should remain active until the user exits the menu. We also do not need "Tabbed into" as it serves the same purpose as Focus. We do not want to overcomplicate the states. Ewa if "active + hover" and "Tabbed into" are needed for other teams, then we will leave it up to Modus to decide.

cjwinsor:
> @cjwinsor @enowak1031 @egunther39 I brought this to the UX team. We are ok with the focus, hover, and active states. We do not need "Active + hover" as the active state does not need to change on hover. It should remain active until the user exits the menu. We also do not need "Tabbed into" as it serves the same purpose as Focus. We do not want to overcomplicate the states. Ewa if "active + hover" and "Tabbed into" are needed for other teams, then we will leave it up to Modus to decide.

@DafneeZambrano, to clarify.  The active + hover state is caused by a user hovering over the element and then interacting with it with a mouse click.  In that moment the styling for active is being combined with the style for hover.  At the point the mouse click starts it should be active and override the hover. From a UX perspective there is no need to detail out this interaction as it makes sense that active should override the hover style.  

As for tabbed into, this was just to illustrate the styling, which in this case was as expected. I think I might have messed up my screenshot for the notification icon, as doing that test again has it match focus... which the original difference was what lead me to include it in the first place.

cjwinsor (approved):

Issue #1558, opened 2023-07-20
Add buttons group component


msankaran0712:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have read the [Modus style guidelines](https://modus.trimble.com/components/web/introduction/)

### Proposal

Created from https://github.com/trimble-oss/website-modus.trimble.com/issues/656

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

Medium

### What product/project are you using Modus Components for ?

modus

### What is your team/division name ?

modus

### Are you willing to contribute ?

None

msankaran0712:
Wait for Styleguide and Figma

msankaran0712:
@Gnirps00 can you please share with us the details about your division name and team/product name?

Gnirps00:
Hi, I'm in Structures division and in SuDO team in Finland.

msankaran0712:
> Hi, I'm in Structures division and in SuDO team in Finland.

Thank you @Gnirps00

msankaran0712:
Assign this to Sprint 71

enowak1031:
@soniakaukonen Do we have a submission for a toggle button? Your links show a button group--different.

soniakaukonen:
@enowak1031 - we renamed it together at some point as button group when looking up the term on Bootstrap.

msankaran0712:
Planned for 2024 Q1

msankaran0712:
Guidelines for the component has to be present on Modus Styleguide and Figma before starting the work on Web components

cc @ElishaSamPeterPrabhu @rthanga1

rthanga1:
@enowak1031 Please attach the figma link here

enowak1031:
https://www.figma.com/file/wyfVJUHWRMkeCfdB38HFEE/Modus---Web?type=design&node-id=24902-8726&mode=design&t=BfJFiRI1DNz59y4q-0

PR #2110, opened 2024-02-02
Issue 1558/button group

ElishaSamPeterPrabhu:
## Description

References #
#1558 

## Type of change

- [ ] Documentation update

## Checklist

### DIFF
--- a/stencil-workspace/src/components/modus-buttongroup/modus-button-group.constants.ts
+++ b/stencil-workspace/src/components/modus-buttongroup/modus-button-group.constants.ts
@@ -1,0 +1,9 @@
+export const OUTLINE_STYLE = 'outline';
+export const FILL_STYLE = 'fill';
+export const PRIMARY_VARIANT = 'primary';
+export const SECONDARY_VARIANT = 'secondary';
+export const SINGLE_SELECT_TYPE = 'single';
+export const DEFAULT_SELECT__TYPE = 'default';
+export const LEFT_BUTTON_POSITION = 'left';
+export const RIGHT_BUTTON_POSITION = 'right';
+export const CENTER_BUTTON_POSITION = 'center';
--- a/stencil-workspace/src/components/modus-buttongroup/modus-button-group.models.tsx
+++ b/stencil-workspace/src/components/modus-buttongroup/modus-button-group.models.tsx
@@ -1,0 +1,19 @@
+import {
+  OUTLINE_STYLE,
+  FILL_STYLE,
+  PRIMARY_VARIANT,
+  SECONDARY_VARIANT,
+  SINGLE_SELECT_TYPE,
+  DEFAULT_SELECT__TYPE,
+  LEFT_BUTTON_POSITION,
+  RIGHT_BUTTON_POSITION,
+  CENTER_BUTTON_POSITION,
+} from './modus-button-group.constants';
+
+export type ButtonGroupStyle = typeof OUTLINE_STYLE | typeof FILL_STYLE;
+export type ButtonGroupVariant = typeof PRIMARY_VARIANT | typeof SECONDARY_VARIANT;
+export type ButtonGroupSelectionType = typeof SINGLE_SELECT_TYPE | typeof DEFAULT_SELECT__TYPE;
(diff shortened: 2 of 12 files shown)


jmwood330t (approved):
Good tests here - and thanks.
Jim

cjwinsor:
@ElishaSamPeterPrabhu I found this example of a button group i think we can mimic how they handle the rounded borders.
https://github.com/goatui/components/tree/main/src/components/button-group

Also, we need to talk to Ewa about the style.  The guidelines currently only show outline and thats all she wants us to implement... but it makes things interesting as the button group could accept any buttons naturally as they are configured if we wanted to.. mixing and matching styles would be odd but doable.  The alternative being that we can have a different child component... `modus-button-group-button` sounds stupid though.

For now, see if you can manage handling the border radius using css variables like my example.  

As for as the `toggleable` option, still considering that part of it.... I'm wondering if we could leverage `type` and have a `toggle` type.. I'll ask in the meeting tomorrow to see if @coliff has any issues or concerns.

I'd almost want to just have it be

```html
<modus-button-group>
  <modus-button type="button"></modus-button>
  <modus-button type="toggle"></modus-button>
  <modus-button type="button"></modus-button>
</modus-button-group>
```
but this leads to people being able to set random sizes, colors, and styles... which would be stupid to do but they could do it...

With your current method or remapping, it could solve that, but it also adds complexity where we could simply just have a slot and throw the buttons in there and let the user deal with each buttons events individually.

Just putting all my current thoughts down as I haven't updated this PR with any of them yet.

cjwinsor:
@ElishaSamPeterPrabhu We will still need the `selection-type` prop so by default its just regular buttons without the isActive being used and the single mode where only one is active at a time.

stencil-workspace/src/components/modus-buttongroup/modus-button-group.constants.ts
```diff
@@ -0,0 +1,9 @@
+export const OUTLINE_STYLE = 'outline';
```
cjwinsor, line 1: Some of these aren't used anymore

stencil-workspace/src/components/modus-buttongroup/modus-button-group.models.tsx
```diff
@@ -0,0 +1,19 @@
+import {
```
cjwinsor, line 1: Cleanup unused. Those we should keep Selection Type as we should have that property.

stencil-workspace/src/components/modus-buttongroup/modus-button-group.tsx
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  @Listen('slotchange')
+  handleSlotChange(event: Event) {
+    const slot = event.target as HTMLSlotElement;
+    const buttons = slot.assignedElements({ flatten: true }).filter((el) => el.tagName === 'modus-button');
+    buttons.forEach((button: HTMLModusButtonElement) => {
+      button.disabled = this.disabled;
+      button.buttonStyle = 'outline';
+      button.type = 'toggle';
```
cjwinsor, line 33: Should only be type toggle when selection type is single, and we would need to add an event listener to make sure only one button is clicked at a time.  So you can keep track of selectedButtons and set isActive to false on any in that array (multiple select coming later would probably need this to be an array).

ElishaSamPeterPrabhu:
I removed the selection-type previously because I thought we didn't finalize on the selection methods .
For now am going with two types single and the default one.

stencil-workspace/src/components/modus-buttongroup/modus-button-group.tsx
ElishaSamPeterPrabhu: isActive is a state in `modus-button,` so are you saying to modify it from button groups?

stencil-workspace/src/components/modus-buttongroup/modus-button-group.models.tsx
ElishaSamPeterPrabhu: Cleaned up

cjwinsor (approved):

PR #2118, opened 2024-02-05
Tabs: Fixed the missed ids and wrong tabIndex issues

DancuVladTrimble:
modus-tabs component

## Description

Last week we integrated the modus-tabs component in our UI application and noticed two small issues:

- the tabs generated by the modus-tabs component do not have an id in HTML even if there is an id property in the tab configuration. This problem affects our QA to access tabs in e2e tests (in our tests we access the HTML element by id);
- all tabs have the same tabIndex = 0;

References: #2117

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

I ran all the e2e and unit tests from your application and my changes didn t affect any tests(all tests passed).

I tested locally and everything looks good, now each tab has the right id set in the tab configuration, also each tab has an unique tabIndex(0, 1, 2, 3,....) :

image

before my changes:

image

## Checklist

### DIFF
--- a/stencil-workspace/package.json
+++ b/stencil-workspace/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@trimble-oss/modus-web-components",
-  "version": "0.15.0",
+  "version": "0.16.0",
   "description": "Trimble Modus Web Component Library",
   "homepage": "https://modus-web-components.trimble.com/",
   "bugs": {
--- a/stencil-workspace/package.json
+++ b/stencil-workspace/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@trimble-oss/modus-web-components",
-  "version": "0.16.0",
+  "version": "0.15.0",
   "description": "Trimble Modus Web Component Library",
   "homepage": "https://modus-web-components.trimble.com/",
   "bugs": {
--- a/stencil-workspace/src/components/modus-tabs/modus-tabs.tsx
+++ b/stencil-workspace/src/components/modus-tabs/modus-tabs.tsx
@@ -54,7 +54,7 @@ export class ModusTabs {
   }
 
   render(): unknown {
-    const tabs = this.tabs.map((tab: Tab, tabIndex: number) => {
+    const tabs = this.tabs.map((tab: Tab) => {
       return (
         <div
           id={`${tab.id}`}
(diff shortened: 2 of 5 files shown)


stencil-workspace/src/components/modus-tabs/modus-tabs.tsx
```diff
@@ -57,12 +57,12 @@ export class ModusTabs {
     const tabs = this.tabs.map((tab: Tab) => {
       return (
         <div
+          id={`${tab.id}`}
           class={`tab ${tab.active ? 'active' : ''} ${this.classBySize.get(this.size)} ${
             this.fullWidth ? 'resizable' : ''
           } `}
           onClick={() => this.handleTabChange(tab.id)}
-          onKeyDown={(event) => this.handleKeyDown(event, tab.id)}
-          tabIndex={0}>
```
cjwinsor, line 65: Without a tabIndex, the tab won't be able to be tabbed into.

stencil-workspace/src/components/modus-tabs/modus-tabs.tsx
```diff
@@ -57,12 +57,12 @@ export class ModusTabs {
     const tabs = this.tabs.map((tab: Tab) => {
       return (
         <div
+          id={`${tab.id}`}
```
cjwinsor, line 60: I need to double check how we want to handle this.  I need to familiarize myself with the component first, but we have utility functions for generating IDs and I need to check if we desire to have the ID given by the user to map directly to the element or not. I'll follow-up on this point soon.

stencil-workspace/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@trimble-oss/modus-web-components",
-  "version": "0.15.0",
+  "version": "0.16.0",
```
cjwinsor, line 3: You don't need to update the version, we will take care of this when we release.

DancuVladTrimble: Ok, got it

stencil-workspace/src/components/modus-tabs/modus-tabs.tsx
DancuVladTrimble: we need to set an id for each tab base on the tabs configuration:

image

without an id our QA is not able to access from e2e tests these tabs.

stencil-workspace/src/components/modus-tabs/modus-tabs.tsx
DancuVladTrimble: @coliff asked me to delete the tabIndex. Also I tested without this property and everything is working like before.

cjwinsor (approved):

Issue #1976, opened 2023-12-21
Sentiment Scale Accessibility / Usability issues


coliff:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/trimble-oss/modus-web-components/blob/main/CONTRIBUTING.md)

### Describe the issue

- The 'sentiment-scale-container' needs a customizable `aria-label` (users may wish to have something like 'Was this page helpful?' or 'how would you rate this product?'
- Disabled state sentiment scale doesn't disable the buttons within it. Please add `aria-disabled="true"`

### Reduced test cases

Go to: https://modus-web-components.trimble.com/?path=/story/components-sentiment-scale--default&args=disabled:true

The buttons are still actionable (have a hover state). When disabled the div should have `aria-disabled="true"`.

### What operating system(s) are you seeing the problem on?

Windows

### What browser(s) are you seeing the problem on?

Microsoft Edge

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

Medium

### What product/project are you using Modus Components for ?

Modus

### What is your team/division name ?

Modus

### Are you willing to contribute ?

Maybe

coliff:
95% of this issue is fixed/resolved. Only thing remaining I can see is that when the sentiment scale is disabled the icons should have CSS for hover state: `pointer-events: none;` and remove the color change on hover.

PR #2121, opened 2024-02-06
Sentiment Scale: Accessibility issues

HarieshPugazhendhi:
## Description

This Sentiment Scale component addresses accessibility to include  customizable aria-label attributes for improved screen reader support.

References #
#1976 
## Type of change

## How Has This Been Tested?

- In the Sentiment-scale component added a control option for aria-label in stories and typed a custom label in it. 
- Verified the presence and functionality of the aria-label attributes  in the Storybook environment by manually inspecting the rendered HTML.

## Checklist


ElishaSamPeterPrabhu (dismissed):

coliff:
Good PR - thanks! There is a remaining issue with the Storybook canvas page:

- The buttons here (https://modus-web-components.trimble.com/?path=/docs/user-inputs-select--default) are disabled. We need to remove `disabled="false"` from `modus-sentiment-scale`.
image

- Not a problem, but the `aria-disabled="false"` which is on by default is unnecessary. Can it be removed? ``aria-disabled="true"`` should be set when it's disabled, but if it's not disabled then `aria-disabled` attribute should removed altogether.

stencil-workspace/src/components/modus-sentiment-scale/modus-sentiment-scale.tsx
```diff
@@ -79,7 +83,7 @@ export class ModusSentimentScale {
                 class={containerClass}
                 onClick={() => this.handleSentimentClick(key)}
                 onKeyDown={(event) => this.handleKeyDown(event, key)}>
-                <SentimentIconMap icon={iconName} size={`${this.type === THUMB_SENTIMENT_TYPE ? '32' : '24'}`} />
+                <SentimentIconMap icon={iconName} size={`${this.type === 'thumbs' ? '32' : '24'}`} />
```
cjwinsor, line 86: This should still use the constant `THUMB_SENTIMENT_TYPE` for maintainability.

cjwinsor (changes requested):

stencil-workspace/src/components/modus-sentiment-scale/modus-sentiment-scale.tsx
HarieshPugazhendhi: Modified the code to still use the constant for maintainability.

cjwinsor:
@HarieshPugazhendhi Just checking if you meant to push latest changes after that last comment or it was just a confirmation of approach.

HarieshPugazhendhi:
> Good PR - thanks! There is a remaining issue with the Storybook canvas page:
> 
> * The buttons here (https://modus-web-components.trimble.com/?path=/docs/user-inputs-select--default) are disabled. We need to remove `disabled="false"` from `modus-sentiment-scale`. This is important because `disabled="false"` actually just means disabled (i.e. it's the same as `disabled="true"`!)
>   image
> * Not a problem, but the `aria-disabled="false"` which is on by default is unnecessary. Can it be removed? `aria-disabled="true"` should be set when it's disabled, but if it's not disabled then `aria-disabled` attribute should removed altogether.

Hey Coliff,
                The changes that you mentioned to remove disabled="false" from modus-sentiment-scale is complicated. Removing this parameter will affect the component's toggle functionality. Furthermore, other parameters rely on the disabled attribute, making its presence important. So what else could be done either?

coliff (approved):
The Canvas demo has accessibility issues (it's ALWAYS announced as disabled - even is disabled if set as False)... but the main functionality on Docs tab is working good so I'll approve.

stencil-workspace/storybook/stories/components/modus-sentiment-scale/modus-sentiment-scale.stories.tsx
```diff
@@ -36,9 +44,9 @@ export default {
   },
 };
 
-const Template = ({iconsType,
+const Template = ({ariaLabel,iconsType,
 disabled}) => html`
-<modus-sentiment-scale type="${iconsType}" disabled="${disabled}"></modus-sentiment-scale>
+<modus-sentiment-scale  aria-label=${ariaLabel} type="${iconsType}" disabled="${disabled}"></modus-sentiment-scale>
```
cjwinsor, line 49: @HarieshPugazhendhi I think you can update this to `?disabled=$"{disabled}"` to satisfy the need to only add the property when it's true.

Here is how attributes are handled in the html templating
https://lit.dev/docs/templates/expressions/#boolean-attribute-expressions

cjwinsor (approved):

Issue #1476, opened 2023-06-30
Modus Select Component - Add a placeholder text


msankaran0712:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/trimble-oss/modus-web-components/blob/main/CONTRIBUTING.md)

### Describe the issue

Missed a fix in #559 

For details contact @sarafarhat13

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

High

### What product/project are you using Modus Components for ?

modus

### What is your team/division name ?

modus

### Are you willing to contribute ?

None

cjwinsor:
We can do Please select and allow user to change with a prop

PR #2133, opened 2024-02-09
Select: Add placeholder prop

ElishaSamPeterPrabhu:
## Description

- Added an initial option to the existing options list which is selected by default is disabled ,value is empty and is hidden. 
- The text content of the added option is "Please help which is the place holder value by default and can be modified with the placeholder prop.

References #
#1476 

## Type of change

- [x] Documentation update

## Checklist


cjwinsor:
@coliff Can you help validate if there are any accessibility issues with this implementation?

coliff:
Let's make the demo iframe a bit taller so the Custom Placeholder one is visible without needing to scroll down in the iframe.
image

For the placeholder, instead of making it the option `hidden`, can we instead just make it `disabled`. See this demo: https://stackblitz.com/edit/modus-select?file=index.html
That way a user must select a different option when opened.

cjwinsor:
Can you also add 24px padding between the the examples

ElishaSamPeterPrabhu:
> Can you also add 24px padding between the the examples

Have added br that seems enough, should it be 24px?

ElishaSamPeterPrabhu:
> * Let's make the demo iframe a bit taller so the Custom Placeholder one is visible without needing to scroll down in the iframe.
>   image
> * This should have linebreaks to prevent it requiring horizontal scroll - supect Prettier formatting it will fix it:
>   image
> * For the placeholder, instead of making it the option `hidden`, can we instead just make it `disabled`. See this demo: https://stackblitz.com/edit/modus-select?file=index.html
>   That way a user must select a different option when opened.

- Added line breaks, increased height in storybook.
- Removed hidden property for placeholder option

cjwinsor (approved):

Issue #2155, opened 2024-02-15
Table | Column visibility dropdown list causes extra scrolling when too long 


apaddock:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/trimble-oss/modus-web-components/blob/main/CONTRIBUTING.md)

### Describe the issue

Max height is set on the table to support sticky headers. 
When the column visibility is enabled and the list of columns is very long, the list extends beyond the table itself, and causes additional scrolling within the page. 

![image](https://github.com/trimble-oss/modus-web-components/assets/4574058/a17a0208-ce91-4df4-b451-4864e66024ce)

**User Problem**

- Overlooking the Apply button as it is out of the page fold.
- Apply button too difficult to access as it needs long scroll to reach on long list of columns.

**Suggested Solution**
- Update column selection panel MAX-height ~~5% less than table height.
- Fixing the cancel and apply buttons to the bottom of the panel so that they are always visible
- The list of columns with checkboxes will be in the scrollable content of the selection panel.

### Reduced test cases

https://codepen.io/Ashley-Sonntag/pen/zYbeYao

### What operating system(s) are you seeing the problem on?

macOS

### What browser(s) are you seeing the problem on?

Chrome

### What is the issue regarding ?

@trimble-oss/modus-web-components

### What version of npm package are you using ?

latest

### Priority

Medium

### What product/project are you using Modus Components for ?

e-Builder

### What is your team/division name ?

e-Builder

### Are you willing to contribute ?

Yes

### Are you using Modus Web Components in production ?

yes

PR #2223, opened 2024-02-29
Table: column visibility dropdown list too long

rjavier-trimbler:
## Description
-adding some css properties to set a propper height for column visibility 
-new maxHeight  prop for ModusTableToolbar to be able to change this outside.

References #2155 

## Type of change

- [ ] Documentation update

## Checklist


J-Ruiz1:
Validated that the column visibility behaves as expected, evidence can be found here https://e-builderinc.atlassian.net/browse/NAV-519

cjwinsor:
@Roliver-Javier I was going to approve this, but see now that a max height property was added for the menu which I disagree with being needed at this time. Can you remove this change?  I see no reason to add a property for this unless there is a strong demand for this level of flexibility.  

But now that I am thinking about it, we probably want some sensible min height though so it doesn't collapse into nothingness.

rjavier-trimbler:
> @Roliver-Javier I was going to approve this, but see now that a max height property was added for the menu which I disagree with being needed at this time. Can you remove this change? I see no reason to add a property for this unless there is a strong demand for this level of flexibility.
> 
> But now that I am thinking about it, we probably want some sensible min height though so it doesn't collapse into nothingness.

Hi @cjwinsor yeah, the purpose of this one is because we have some views that are using the table with different spaces and when the toolbar collapses it looks too long (the number of columns will determine the height) so in this case we can set the maximum and the content will be scrollable.

dnarey:
Hey @cjwinsor.  Thanks for the feedback.  The addition of a max height property was from our UX team.  I asked if they could please add a summary and justification, which they included here: https://e-builderinc.atlassian.net/browse/NAV-519?focusedCommentId=381716

Please let us know if there is anything else you need, or if we should take a different approach.  Thank you!

cjwinsor:
@Roliver-Javier After some discussion with Teja and @yohernandez and @dnarey, for now lets use the tables `maxHeight` as the `maxHeight` of the menu.  If the table doesn't have a `maxHeight` defined, then use 60dvh (not vh) as the height so it can account for mobile keyboard sorta elements.  This will allow us to postpone adding a new property to the api until we get more feedback to support it as a necessity. This means that the height will be set on the element, vs a css clas.  Let me know if you have any questions.

J-Ruiz1:
Validated that maxHeight is set to 60dvh

image

cjwinsor (approved):

Issue #2242, opened 2024-03-01
Alert: Limit the amount of total characters to 300


cjwinsor:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have read the [Modus style guidelines](https://modus.trimble.com/components/web/introduction/)

### Proposal

Limit the amount of text in the alert to 300 characters, truncate afterwards.

### Motivation and context

Guidelines are going to be updated to reflect this restriction.

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

Medium

### What product/project are you using Modus Components for ?

Modus

### What is your team/division name ?

Modus

### Are you willing to contribute ?

Yes

PR #2268, opened 2024-03-06
Alert: Update max character length to 300

ElishaSamPeterPrabhu:
## Description

- Modified the message string to be a substring of 300 characters

References #

 #2242 

## Type of change

- [ ] Documentation update

## Checklist


stencil-workspace/src/components/modus-alert/modus-alert.tsx
```diff
@@ -73,7 +73,7 @@ export class ModusAlert {
           {this.type === 'warning' ? <IconWarning size={iconSize} /> : null}
         </div>
         <div class="message">
-          {this.message}
+          {this.message?.substring(0, 300)}
```
jmwood330t, line 76: Do we have unit tests that do boundary value analysis, i.e. tests with messages with lengths such as:

- Some number between (1, 298) inclusive
- 299
- 300
- 301
- Some number greater than 301

? And how should it handle the case when the length is greater than 300?

ElishaSamPeterPrabhu: When the text length is above 300 the characters are limited , so the test cases are written to check when the characters given are above 300 and the `div.message`'s text content length is 300 and the message is set to have a text with the text length to be 299 and is checked again to match it to 299.

cjwinsor (approved):

Issue #2272, opened 2024-03-06
Modus Icon buttons: Remove Pressed State When Popup is closed 


Sanket8793:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/trimble-oss/modus-web-components/blob/main/CONTRIBUTING.md)

### Describe the issue

Modus Navbar has its own implementation for icon buttons, the icon button should remains in the active state after the popover is closed. The button is expected to return to the default state after the popover is closed.

### What browser(s) are you seeing the problem on?

Chrome

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

Medium

### What product/project are you using Modus Components for ?

E-Builder/Branding

### What is your team/division name ?

E-Builder/DarkKnights

### Are you willing to contribute ?

Yes

PR #2276, opened 2024-03-07
Navbar: Remove Pressed State for the icon button after popup is closed

Sanket8793:
## Description

Remove Pressed State for the icon button after popup is closed

References #2272 

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

when the popup is closed make sure that icon button is expected to return to the default state.

## Checklist


zabdur-rahman:
@cjwinsor Can this ticket be looked at? It's been sitting for a couple of weeks.

cjwinsor:
@zabdur-rahman @Sanket8793  Sorry about the delay, focus has been on other initiatives and tickets.  This particular PR demands a bit more of my time as I do not agree with the implementation and will need to come up with what I think is appropriate.  I was going to review later today now the previous blockers are resolved, but I'll review now.

A prop is added for `isDynamicButtonSlotOpen` which represents that someone could try to set this to `true` but that would have no affect as its not used in that capacity in the code.  I could not find any value in making it a public facing prop.    

In the `showButtonMenuById(id)` function, its setting this to true if `this.openButtonMenuId === id`.   Then it calls the function to `hideMenus()` which will set `this.openButtonMenuId = undefinded`.  The code continues to check if `this.openButtonMenuId !== id && !this.isDynamicButtonSlotOpen` before setting the `this.openButtonMenuId = id`.  Since `this.openButtonMenuId` was set to `undefined` in the previous step, the first condition will pass as long as the passed in `id` is not undefined.  The second condition will pass when whatever the previous state of `this.openButtonMenuId !== id`.  

If `this.isDynamicButtonSlotOpen` was a `const` inside of `showButtonMenuById` i don't think it changes how the code operates.  Also you're setting a prop without having it configured as `mutable: true` which means it will throw a warning when its updated inside the code as done here.  

Currently I don't know how to reproduce even when I try in the e-Builder QA3 environment it already removes the active state when I click elsewhere on the page, though the userway component never opens when I do click on the menu item. The "How Has This Been Tested?" area is the best spot to give details which enable me to quickly validate the scenario.  Screenshots, gifs, videos, code snippets all help to reduce the effort to review.  Both the issue writeup and the PR writeup are vague, leading me to save it for when I have time to review it thoroughly. In the future, I will try to indicate why PRs are not being prioritized so you have the opportunity to make adjustments as needed.

Can you add to this and future PRs:
- Reproduction case I can follow.
- Explanation of what the issue was
- Explanation on how this code change fixes the issue

cjwinsor:
@Sanket8793 Thank you for the explanation and the video, I'm able to see the issue now.  The main code change I was looking for was to remove the `isDynamicButtonSlotOpen` prop `altogether` and simply use a `const` variable inside `showButtonMenuById`.

Sanket8793:
> @Sanket8793 Thank you for the explanation and the video, I'm able to see the issue now. The main code change I was looking for was to remove the `isDynamicButtonSlotOpen` prop `altogether` and simply use a `const` variable inside `showButtonMenuById`.

@cjwinsor I updated the code using `const` variable inside `showButtonMenuById`.

cjwinsor (approved):

Issue #2347, opened 2024-03-25
Text-input: Add attributes (`autocorrect`, `autocapitalize`, `enterkeyhint`, `spellcheck`)


coliff:
We should add missing attributes to the text-input and textarea

- [autocapitalize](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize)
- [autocorrect](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text#autocorrect) (this is non-standard, but can be important for mobile)
- [Enterkeyhint](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/enterkeyhint)
- [spellcheck](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text#spellcheck)

PR #2351, opened 2024-03-25
Input: Add `autocapitalize`, `autocorrect`, `enterkeyhint`, and `spellcheck `properties

austinoneil:
## Description

Add autocapitalize, autocorrect, enterkeyhint, and spellcheck properties to input

References #2347

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

Added unit tests

## Checklist

### DIFF
--- a/stencil-workspace/src/components.d.ts
+++ b/stencil-workspace/src/components.d.ts
@@ -1313,9 +1313,17 @@ export namespace Components {
          */
         "autoFocusInput": boolean;
         /**
+          * (optional) Capitalization behavior when using a non-traditional keyboard (e.g. microphone, touch screen)
+         */
+        "autocapitalize": boolean | 'none' | 'off' | 'sentences' | 'on' | 'words' | 'characters';
+        /**
           * (optional) Sets autocomplete on the input.
          */
         "autocomplete": string | null;
+        /**
+          * (optional) Whether to activate automatic correction while the user is editing this field in Safari.
+         */
+        "autocorrect": boolean | 'off' | 'on';
         /**
           * (optional) Whether the input has a clear button.
          */
@@ -1324,6 +1332,10 @@ export namespace Components {
           * (optional) Whether the input is disabled.
          */
         "disabled": boolean;
+        /**
+          * (optional) Which action label to present for the enter key on virtual keyboards.
+         */
+        "enterkeyhint": 'enter' | 'done' | 'go' | 'next' | 'previous' | 'search' | 'send';
         /**
           * (optional) The input's error state text.
          */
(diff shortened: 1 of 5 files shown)


coliff (changes requested):
Nice! looks good.. but we need to have the additional controls to the demo canvas:
https://deploy-preview-2351--moduswebcomponents.netlify.app/?path=/story/user-inputs-text-input--default
(`autocapitalize`, `autocorrect`, `enterkeyhint`, and `spellcheck` options should be available here)

I don't think we need demos of them on the Docs page, can you please remove?

austinoneil:
@coliff done.

coliff:
- autocapitalize is appearing twice
- spellcheck is missing
- changing the values of the autocapitalize, autocorrect and enterkeyhint does not change the input at all. See screenshot:
image
We expect the input with id="mwc_id_0_text_input" here to have enterkeyhint="done" autocapitalize="sentences" autocorrect="off"

austinoneil:
Oops.  Fixed.

coliff:
thanks for work on this so far - still a few more things:
- `autocapitalize` by default should not be set at all. autocapitalize attribute should not be set on the `mwc_id_9_text_input` in the canvas example. (autocapitalize is the default on mobile devices usually anyway). So the dropdown in the Controls as standard should be 'Choose option...'
- `autocorrect` by default should not be set at all. autocorrect attribute should not be set on the `mwc_id_9_text_input` in the canvas example. (autocorrect is the default on mobile devices usually anyway). So the dropdown in the Controls as standard should be 'Choose option...'
- `enterkeyhint` should not have an option pre-selected - so have a ('Choose option...') selected.
- `spellcheck` should be a dropdown with true or false options - but by default nothing selected ('Choose option...').

austinoneil:
I made the changes (except for spellcheck, as that is an attribute that is `true` if it's included, much like `disabled`).

coliff (approved):

coliff (approved):

kuhnboy:
@austinoneil Why make autocorrect: boolean | 'off' | 'on'; 

Why not stick to just an enumeration or just a boolean?

austinoneil:
@kuhnboy  When I as a developer see `autocorrect` as an attribute, my instinct tells me that it is a `boolean` input; however, the [`autocorrect`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text#autocorrect) attribute on `input` takes `"on"` and `"off"` as values.  I included both because doing so makes the library easier to use without creating any obvious problems.

austinoneil:
@kuhnboy That said, it does look like this was causing issues with `autocapitalize` in react (see #2412), so if needed, we can just use `"on"` and `"off"` (or just `boolean`)

Issue #2076, opened 2024-01-26
Modal Accessibility Issues


coliff:
The Modal Component has a few accessibility issues:

- A closed/unopened Modal should not have `role="dialog"`, that should be added when modal is opened and removed when modal is closed.
- Modal should have `aria-modal="true"` when opened (removed when closed)
- Modal should not have `aria-label` when closed.
- Modal should have `aria-hidden="true"` when closed
- Right now, the `aria-label` has to manually set and is listed as 'optional'. I think it'd be better if the value of the modal title was used automatically. The title would have an id (e.g. `id="contactFormModalTitle"`, and the modal would then have something like `aria-labelledby="contactFormModalTitle"` then no need for aria-label to be used at all (this is what Bootstrap does)

PR #2374, opened 2024-03-29
Modal: Accessibility updates

austinoneil:
## Description

* When not visible, role is not "dialog"
* When visible, has "aria-modal" attribute
* When not visible, "aria-label" is not present
* When not visible, "aria-hidden" is true

References #2076 (note: I am leaving the final bullet point off, as that is a breaking change and I am not an employee/owner and there seems to be ambiguity as to whether or not we should implement that change.)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

E2E tests

## Checklist

### DIFF
--- a/stencil-workspace/src/components/modus-modal/modus-modal.e2e.ts
+++ b/stencil-workspace/src/components/modus-modal/modus-modal.e2e.ts
@@ -22,6 +22,82 @@ describe('modus-modal', () => {
     expect(element.innerText).toContain('New Text');
   });
 
+  it('does not have the role of "dialog" when not visible', async () => {
+    const page = await newE2EPage();
+
+    await page.setContent('<modus-modal></modus-modal>');
+    const element = await page.find('modus-modal >>> div');
+    expect(element.getAttribute('role')).not.toEqual('dialog');
+  });
+
+  it('has the role of "dialog" when not visible', async () => {
+    const page = await newE2EPage();
+
+    await page.setContent('<modus-modal></modus-modal>');
+    const component = await page.find('modus-modal');
+    const element = await page.find('modus-modal >>> div');
+    await component.callMethod('open');
+    await page.waitForChanges();
+    expect(element.getAttribute('role')).toEqual('dialog');
+  });
+
+  it('does not have aria-modal attribute when not visible', async () => {
+    const page = await newE2EPage();
+
+    await page.setContent('<modus-modal></modus-modal>');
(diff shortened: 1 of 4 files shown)


coliff (approved):

coliff:
this is great - thanks!
https://deploy-preview-2374--moduswebcomponents.netlify.app/?path=/docs/components-modal--default

Can see the DOM of the modal when closed is:
`<div class="modus-modal overlay hidden" style="z-index: 1;" aria-hidden="true">`
when open it is:
`<div class="modus-modal overlay visible" style="z-index: 1;" aria-label="Modal" aria-modal="true" role="dialog">`

Exactly what we needed 👍

coliff (changes requested):
Oops, found an accessibility bug in the Canvas demo - aria-label is in the DOM even if no value is provided:
image

austinoneil:
I am seeing the same issue in [prod's storybook.](https://modus-web-components.trimble.com/?path=/story/components-modal--default&args=ariaLabel:). As is, it looks like a blank string is being treated like a presence of an `aria-label`.  I'm not sure whether this would be considered a bug/issue.

coliff:
> I am seeing the same issue in [prod's storybook.](https://modus-web-components.trimble.com/?path=/story/components-modal--default&args=ariaLabel:). As is, it looks like a blank string is being treated like a presence of an `aria-label`. I'm not sure whether this would be considered a bug/issue.

It is a problem. The Modal causes a violation in the Accessibility panel. We want to get all them resolved. There should never be any `aria-label=""` anywhere in any of the code in any of the components. Really, this is an issue we should resolve in all the Modus Web Components...

Overall, the changes in this PR are an improvement so I'm thinking we should merge as it is and look at improving further in a different issue/PR. For now, perhaps we just remove the 'Modal' value from `aria-label` in the Storybook Canvas. That way there will no accessibility issues reported at all in Storybook for the Modal component.
Demo: https://deploy-preview-2374--moduswebcomponents.netlify.app/?path=/story/components-modal--default&args=ariaLabel:

austinoneil:
@coliff I have filed  #2409

coliff:
> For now, perhaps we just remove the 'Modal' value from `aria-label` in the Storybook Canvas. That way there will no accessibility issues reported at all in Storybook for the Modal component.

I pushed this change and will merge. thanks! 👍

coliff (approved):

Issue #2212, opened 2024-02-27
Navbar: navbar `<nav>` should have an `aria-label`


coliff:
Navbar needs to have an optional `aria-label` config.

Reason being is that if there is more than one `<nav>` component on the page then we want screen reader to announced that the navbar is 'main menu' or something like that.

PR #2379, opened 2024-03-30
Navbar: Add `aria-label` attribute

austinoneil:
## Description

Add `aria-label` attribute to navbar.

References #2212

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

e2e tests, inspect element on storybook

## Checklist


stencil-workspace/src/components/modus-navbar/modus-navbar.tsx
```diff
@@ -364,7 +367,7 @@ export class ModusNavbar {
 
     return (
       <Host id={this.componentId}>
-        <nav class={`${direction} ${shadow} ${variant}`}>
+        <nav class={`${direction} ${shadow} ${variant}`} aria-label={this.ariaLabel}>
```
coliff, line 370: If no aria-label is set then it renders as `<nav class aria-label>` - however it shouldn't show aria-label at all if not specified. Can you fix that please? By default it shouldn't have an aria-label at all.

coliff (changes requested):

stencil-workspace/src/components/modus-navbar/modus-navbar.tsx
austinoneil: I was not able to reproduce this locally.  When running locally, I got desired behavior.  I added some unit tests for these user journeys.

coliff: image

Adding `aria-label` to the modus-navbar is problematic, I think it should be on the nav only.

austinoneil:
I changed the prop name to `navAriaLabel` and update documentation/tests.

coliff:
It's still rendering empty aria-label if no aria-label is specified but it shouldn't. See code here:

image

coliff:
Hi @austinoneil - the outstanding issue on this has been resolved now with your merged PR (empty aria-label fix) so once the conflicts are resolved here we can merge too. thanks :-)

coliff:
doesn't seem to work:
https://deploy-preview-2379--moduswebcomponents.netlify.app/?path=/story/components-navbar--default&args=ariaLabel:default
This link should give the navbar an aria-label with default value but no value is given... its just rendered as `<nav class="" aria-label="">`

austinoneil:
@coliff fixed.

coliff (approved):

PR #2380, opened 2024-04-01
Tooltip:focus tooltipped item when tooltip is active, allow tooltip to be dismissed on escape

austinoneil:
## Description

Focus tooltipped item when tooltip is active, allow tooltip to be dismissed on escape

References  #2177

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

End to end tests, manual test using the following html:

```
    <modus-tooltip text="Tooltip text..." position="right">
      <modus-button tabindex="3">Button</modus-button>
      <modus-button>Button2</modus-button>
    </modus-tooltip>
```

Also tried without `tabindex` on first button.

## Checklist

### DIFF
--- a/stencil-workspace/src/components/modus-tooltip/modus-tooltip.tsx
+++ b/stencil-workspace/src/components/modus-tooltip/modus-tooltip.tsx
@@ -1,5 +1,5 @@
 // eslint-disable-next-line
-import { Component, Element, Fragment, h, Prop, Watch } from '@stencil/core';
+import { Component, Element, h, Prop, Watch, Fragment } from '@stencil/core';
 import { createPopper, Instance } from '@popperjs/core';
 import { ModusToolTipPlacement } from './modus-tooltip.models';
 
@@ -47,6 +47,7 @@ export class ModusTooltip {
   }
 
   private popperInstance: Instance;
+  private targetTabIndex?: string; // necessary for preserving tab index after blur
   private tooltipElement: HTMLDivElement;
   private readonly showEvents = ['mouseenter', 'mouseover', 'focus'];
   private readonly hideEvents = ['mouseleave', 'blur', 'click'];
@@ -72,6 +73,8 @@ export class ModusTooltip {
     const target = this.element.firstElementChild;
     if (!target || !this.tooltipElement) return;
 
+    this.targetTabIndex = target.getAttribute('tabindex');
+
     this.popperInstance = createPopper(target, this.tooltipElement, {
       placement: position,
       modifiers: [
@@ -91,6 +94,8 @@ export class ModusTooltip {
     this.hideEvents.forEach((event) => {
       target.addEventListener(event, this.hideEventsListener);
     });
+
+    target.addEventListener('keydown', this.escapeKeyHandler);
   }
 
   cleanupPopper(): void {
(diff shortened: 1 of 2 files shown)


stencil-workspace/src/components/modus-tooltip/modus-tooltip.tsx
```diff
@@ -109,7 +114,17 @@ export class ModusTooltip {
...
+    if (event.code === 'Escape') {
+      this.hide();
+    }
+  };
+
   show(): void {
+    const target = this.element.firstElementChild as HTMLElement;
+    const tabIndex = this.targetTabIndex !== '' && this.targetTabIndex != null ? (this.targetTabIndex as string) : '-1';
+    target.setAttribute('tabindex', tabIndex);
+    target.focus();
```
austinoneil, line 127: Focusing for two reasons:
1. According to [ARIA standards](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/#keyboardinteraction), focus stays on the triggering element while the tooltip is displayed.
2. Keyboard interaction is done on focused item only.

coliff (dismissed):

cjwinsor:
@austinoneil Is it possible that this could be simplied to just adding something like 

```tsx
  @Listen('keyup')
  escapeKeyHandler(event: KeyboardEvent) {
    if (event.code === 'Escape') {
      this.hide();
    }
  }
```

The idea being that `@Listen` will listen for the `keyup` event on the Host, so esc won't affect tooltip element unless the focus was on an item within the tooltip and bubbled up. 

This avoids `tabIndex` handling, which I always try to avoid.  Let me know if I am overlooking some scenario.  

Also I used `keyup` instead of `keydown` as it would fire once at the end, while `keydown` will continue to fire over and over again if held.  It occurs to me now that many, if not all, of our other components might be needing this consideration in the future.  

FYI @coliff

austinoneil:
> @austinoneil Is it possible that this could be simplied to just adding something like
> 
> ```tsx
>   @Listen('keyup')
>   escapeKeyHandler(event: KeyboardEvent) {
>     if (event.code === 'Escape') {
>       this.hide();
>     }
>   }
> ```
> 
> The idea being that `@Listen` will listen for the `keyup` event on the Host, so esc won't affect tooltip element unless the focus was on an item within the tooltip and bubbled up.
> 
> This avoids `tabIndex` handling, which I always try to avoid. Let me know if I am overlooking some scenario.
> 
> Also I used `keyup` instead of `keydown` as it would fire once at the end, while `keydown` will continue to fire over and over again if held. It occurs to me now that many, if not all, of our other components might be needing this consideration in the future.
> 
> FYI @coliff

The issue with this is that `keyevent`s only occur on the focused element, which needs a non-negative `tabindex`.  The reasoning for this requirement is that having more than one element with the same keyevent listener can cause issues (imagine hitting "enter" and suddenly, every button were triggered.)

I'll file an issue for replacing `keydown` events with `keyup` events.  I see the issue with using `keydown`.

prashanth-offcl (changes requested):
@austinoneil I tested the implementation with multiple buttons with tooltips locally and have a few observations:

- I need to press the tab key twice to navigate between buttons with tooltips.
- Remove listener in cleanupPopper fn.

I understand the issue with using @Listen since you have to specify the target as document, body, or window because the tooltip is not focusable. Since you are adding the listener to the parent, the problem of all tooltip instance listeners triggering should not occur. Therefore I believe we should be able to remove the tabIndex logic, can you try it?

Issue #1977, opened 2023-12-21
Add `aria-sort` attribute for column sorting


coliff:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/trimble-oss/modus-web-components/blob/main/CONTRIBUTING.md)

### Describe the issue

It'd be good to add `aria-sort` attribute to table headers when using column sort.

Some details here:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort

### What operating system(s) are you seeing the problem on?

Windows

### What browser(s) are you seeing the problem on?

Microsoft Edge

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

Low

### What product/project are you using Modus Components for ?

Modus

### What is your team/division name ?

Modus

### Are you willing to contribute ?

No

PR #2425, opened 2024-04-09
Table: Add aria-sort to header

ElishaSamPeterPrabhu:
## Description

- Added aria-sort to table header
References #
Fixes: #1977 
## Type of change

- [ ] Documentation update

## Checklist

### DIFF
--- a/stencil-workspace/src/components/modus-table/parts/columnHeader/modus-table-column-header.tsx
+++ b/stencil-workspace/src/components/modus-table/parts/columnHeader/modus-table-column-header.tsx
@@ -2,7 +2,7 @@ import {
   FunctionalComponent,
   h, // eslint-disable-line @typescript-eslint/no-unused-vars
 } from '@stencil/core';
-import { Header } from '@tanstack/table-core';
+import { Column, Header } from '@tanstack/table-core';
 import { KEYBOARD_ENTER } from '../../modus-table.constants';
 import { ModusTableColumnResizingHandler } from './modus-table-column-resizing-handler';
 import { ModusTableColumnHeaderLabel } from './modus-table-column-header-label';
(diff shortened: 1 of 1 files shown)


coliff (changes requested):
is isn't working as expected yet. Columns which are not sorted are being announced as `aria-sort="ascending"`. Columns should only have that if they are sorted ascendingly.

For example here- the first name column is not sorted so it shouldn't have aria-sort attribute at all.
image

stencil-workspace/src/components/modus-table/parts/columnHeader/modus-table-column-header.tsx
```diff
@@ -25,7 +25,17 @@ interface ModusTableColumnHeaderProps {
...
+  return isColumnResizing
+    ? '' // When column resize is enabled, we don't show the tooltip.
+    : column.getIsSorted() === 'asc'
+      ? 'ascending'
+      : column.getIsSorted() === 'desc'
+        ? 'descending'
+        : column.getNextSortingOrder() === 'asc'
+          ? 'ascending'
+          : 'descending';
+}
```
cjwinsor, line 38: This logic could be more clear with
```js
if (isColumnResizing) return '';

const currentSort = column.getIsSorted();
if (currentSort === 'asc') return 'ascending';
if (currentSort === 'desc') return 'descending';

const nextSort = column.getNextSortingOrder();
return nextSort === 'asc' ? 'ascending' : 'descending';
```

ElishaSamPeterPrabhu:
> It isn't working as expected yet. Columns which are not sorted are being announced as `aria-sort="ascending"`. Columns should only have that if they are sorted ascendingly.
> 
> For example here- the first name column is not sorted so it shouldn't have aria-sort attribute at all. image

Updated logic

coliff (approved):

cjwinsor (approved):

PR #2450, opened 2024-04-18
Switch: Set aria-checked value to "true" and "false"

austinoneil:
## Description

Previous implementation was passing a boolean in as the attribute value.  New implementation passes in a string.

References  #2143

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?

Add e2e checks, update unit tests

https://deploy-preview-2450--moduswebcomponents.netlify.app/?path=/story/user-inputs-switch--medium

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #2455, opened 2024-04-20
Completely remove `aria-checked` from switch, set `id` on `input` to be the same as `label`, replace brittle unit test with more precise e2e test

austinoneil:
… behavior

## Description

Completely remove `aria-checked` from switch, set `id` on `input` to be the same as `label`, replace brittle unit test with more precise e2e test

References https://github.com/trimble-oss/modus-web-components/pull/2450

## Type of change

- [ ] Documentation update

## Checklist

### DIFF
--- a/stencil-workspace/src/components/modus-switch/modus-switch.tsx
+++ b/stencil-workspace/src/components/modus-switch/modus-switch.tsx
@@ -69,11 +69,11 @@ export class ModusSwitch {
           <span class={`slider`}></span>
         </div>
         <input
-          aria-checked={String(this.checked)}
           aria-disabled={this.disabled ? 'true' : undefined}
           aria-label={this.ariaLabel}
           checked={this.checked}
           disabled={this.disabled}
+          id={this.label}
           ref={(el) => (this.checkboxInput = el as HTMLInputElement)}
           role="switch"
           type="checkbox"></input>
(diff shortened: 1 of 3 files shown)


coliff (changes requested):
This will be great when it's finished - but it's not quite right yet.
image

- The label is not matching the id.
- When the switch is checked it should have the `checked` attribute
- id name should assigned automatically with something like `mwc_id_0_switch_input` like do for text-inputs (`<label for="mwc_id_0_text_input">Username</label>`)

austinoneil:
> When the switch is checked the input should have the checked attribute
This issue is also present in `main`.  It is definitely strange, as the `checked` attribute is added to the `input` element.  I think this should be filed as an issue.

I've fixed the other two issues.

stencil-workspace/src/components/modus-switch/modus-switch.tsx
```diff
@@ -1,4 +1,5 @@
 // eslint-disable-next-line
+import { generateElementId } from '../../utils/utils';
 import { Component, Prop, h, Event, EventEmitter, Listen } from '@stencil/core';
```
coliff, line 3: I think this needs a `// eslint-disable-next-line`

coliff:
oh, found a bug. Clicking on the label should change the state like this:
https://modus.trimble.com/components/web/switches/#overview

Issue #2282, opened 2024-03-13
Autocomplete Behavior


enowak1031:
https://trimble-arch.aha.io/ideas/ideas/MOD-I-35

1.) New flavor of autocomplete that shows all items when opened (rather than only showing items based on user input). I believe this needs to be a flavor only for the use cases when the list within the autocomplete is way too long. a.) it would confuse the user why everything is showing without knowing any context, and b,) it could take forever to load a potenially irrelevant list. 

2.) Scroll the list to the selected item (to assure user of their selection): once the user selects something from the autocomplete list, and reopens the autocomplete, they should see their selection within the context of the list. The already selected item item shows with a checkmark on the right and is the second one from the bottom of the list. 

**This is going to be addressed in a separate issue:** 
3.) If this is a multliselect autocomplete: show chips ABOVE the input field (not inside). When chips are created, and the field is reopened, the first selection in alphabetical order should be visible as the second from the bottom (unless it's the first one in the list or the last one in the list). 

https://github.com/trimble-oss/modus-web-components/issues/2446

apaddock:
@cjwinsor or @enowak1031 when will this be approved for engineering?

enowak1031:
@apaddock This is approved and ready to be worked on! I broke the story up to exclude #3 on the list. But you can break it up further, if need be.

cjwinsor:
@apaddock Discussed with team, and these requirements are approved to be implemented.

PR #2484, opened 2024-04-26
Autocomplete: show all items Behavior

rjavier-trimbler:
## Description

References #2282  #2567 

This PR without knowing also fixes the issue:: #2567  

## Type of change

- [ ] Documentation update

## Checklist


stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
```diff
@@ -195,7 +209,14 @@ export class ModusAutocomplete {
   };
 
   handleInputBlur = () => {
-    this.hasFocus = !this.disableCloseOnSelect;
+    if (!this.hasOptionMatched(this.value)) {
```
cjwinsor, line 212: Nesting if statements makes the code harder to read.  Can you switch this to something like 
```js
if (this.hasOptionsMatched(this.value) || 
    this.selectedOption === '' || 
    this.value === '' || 
    this.hasFocus) {
  return;
}

this.value = this.selectedOption;
this.disableFiltering = true;
```

This makes it easier to tell what conditions cause it to not continue.

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
```diff
@@ -321,8 +343,11 @@ export class ModusAutocomplete {
     search = search || '';
     const isSearchEmpty = search.length === 0;
 
-    if ((isSearchEmpty && !this.showOptionsOnFocus) || (this.disableFiltering && this.disableCloseOnSelect)) {
+    if (isSearchEmpty) this.selectedOption = '';
```
cjwinsor, line 346: one liners generally make code harder to read as it puts the action at a variable distance from the left depending on the condition.  Update to: 

```js
if (isSearchEmpty) {
  this.selectionOption = '';
}
```

cjwinsor (changes requested):
I haven't functionally validated this works.  Has this been QA'd by your team?

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
rjavier-trimbler: done!

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
rjavier-trimbler: done!

J-Ruiz1:
@cjwinsor this has already been tested for the following scenarios [in this user story](https://e-builderinc.atlassian.net/browse/NAV-589):

**Scenario 1**: Validate that the Process Filter shows all available options when focused

**Scenario 2**: Validate that the Process Filter is able to filter based on search criteria via adding characters

**Scenario 3**: Validate that the Process Filter is able to filter based on search criteria via removing characters. An empty again Process filter must show all available options

**Scenario 4**: Validate that the Process Filter clears all characters and show the placeholder text "search processes" when characters are entered but the filter loses focus without a selection

**Scenario 5**: Validate that the Process Filter shows all available options when refocusing on the filter after an option was previously selected. The already selected item item shows with a checkmark on the right and is the second one from the bottom of the list  (when able)

**Scenario 6**: Validate that when there are no matches when searching with no previous selection the dropdown will show a “not found” message.

**Scenario 7**: Validate that when there are no matches when searching with a previous selection the dropdown will show a “not found” message. Focusing out will show the previously selected option

**Scenario 8**: Validate that clearing the autocomplete removes the highlighted selection.

cjwinsor (changes requested):
Things I noticed:

A flicker was introduced when picking an item (in my case, I didn't set any props, only set `options`) that wasn't there before. The text goes blank, then shows new value.

It now behaves as if `show-options-on-focus` is always enabled.  It should not show the items on focus until that prop is set to true.  There seems to be a previous bug where once there is a selection it opens the autocomplete on focus even when `show-options-on-focus` is not set to `true`, that isn't correct either.  

If I have a value selected, and while focused on the textbox, click the down arrow, it erases the text and the selection.  The down arrow should not remove the value/selection.

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
```diff
@@ -365,6 +389,22 @@ export class ModusAutocomplete {
...
   }
 
+  scrollToOptionSelected = () => {
+    if (this.multiple || this.focusItemIndex === 0) {
+      return;
+    }
+    const optionList = this.el.shadowRoot.querySelector(`.options-container`) as HTMLUListElement;
+
+    const selectedOption = optionList.querySelector('li.selected') as HTMLElement;
+    if (selectedOption) {
```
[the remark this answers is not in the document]
rjavier-trimbler: I did but there's a requirement that says 

`The already selected item item shows with a checkmark on the right and is the second one from the bottom of the list`.

So I had to play around with the offset position to achieve that.  I added it anyway.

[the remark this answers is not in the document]
cjwinsor: Why is there such a super specific requirement that "is the second one from the bottom"... seems arbitrary and over complicates.  The key is that the item is visible. if its the first x number of items, it can never be second from bottom... and if it is the last item as well... very arbitrary and I don't agree with that aspect of this.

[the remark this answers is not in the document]
rjavier-trimbler: the requirement is this:



This autocomplete has 16 options:

If i click on the option 10 which it cant be displayed because of the max-height then the scroll position will be moved to show the selected option as the second from bottom only if this is not the last item. This wont happen if you select the first option or any of the options that are displayed without the help of scrolling.
Screen Shot 2024-06-07 at 11 53 54 AM

Screen Shot 2024-06-07 at 11 59 25 AM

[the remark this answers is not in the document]
cjwinsor: @rjavier-trimbler I understood the requirements, I just didn't agree with them. I will have to talk to UX on Monday on this one.  There are no modus guidelines around where the selection is positioned and this particular aspect was not in the Aha submission that was approved.  It increases the complexity of the code with no basis of support from the Modus Design System.  My issue is you were given this as a requirement when it was outside the scope of what was approved.

J-Ruiz1:
Validated that the autocomplete complies with the following scenarios

Scenario 1: Validate that the Process Filter shows all available options when focused

Scenario 2: Validate that the Process Filter is able to filter based on search criteria via adding characters

Scenario 3: Validate that the Process Filter is able to filter based on search criteria via removing characters. An empty again Process filter must show all available options

Scenario 4: Validate that the Process Filter shows all available options when refocusing on the filter after an option was previously selected. 

Scenario 5: Validate that when there are no matches when searching with no previous selection the dropdown will show a “not found” message.

Scenario 6: Validate that when there are no matches when searching with a previous selection the dropdown will show a “not found” message. 

Scenario 7: Validate that clearing the autocomplete removes the highlighted selection.

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
```diff
@@ -284,6 +293,14 @@ export class ModusAutocomplete {
     }
   }
 
+  handleTextInputFocus = () => {
+    console.log(this.value);
```
cjwinsor, line 297: Can you remove this console.log

cjwinsor (changes requested):

cjwinsor (approved):

PR #2509, opened 2024-05-08
Content Tree: Event trigger on shift arrow keys

prashanthr6383:
## Description

Created a function to handle click event using <kbd>shift</kbd> + <kbd>arrow  keys</kbd> 

Fixes #1713

## Type of change

- [ ] Documentation update

## Checklist


cjwinsor:
@prashanthr6383 Can you take a look to make sure that the `itemClick` event is fired for the correct element when its selected and deselected via the shift arrow keys?  If I add a listener to a single element (Social) and i select the item above it first, then do shift down, it doesn't fire the event, but when i unselect it, it does with shift up.  It also fires on shift down, which shoudl trigger the item click for the below item, not Social.  Double check the arg that is returned as well: True for selected, False for unselected

prashanthr6383:
> @prashanthr6383 Can you take a look to make sure that the `itemClick` event is fired for the correct element when its selected and deselected via the shift arrow keys? If I add a listener to a single element (Social) and i select the item above it first, then do shift down, it doesn't fire the event, but when i unselect it, it does with shift up. It also fires on shift down, which shoudl trigger the item click for the below item, not Social. Double check the arg that is returned as well: True for selected, False for unselected

@cjwinsor Updated the code. Now the event trigger for shift up and shift down will return true for selected and false for unselected, but though the `itemClick` seems to be firing correctly, the selected element is the previous one. Can you confirm once if this is the expected behavior of shift + arrow or still needs to be fixed?

prashanth-offcl (changes requested):
I see a lot of duplicate events emitted while selecting an option, please check.

prashanthr6383:
> I see a lot of duplicate events emitted while selecting an option, please check.

I am not able to reproduce it. may be a storybook issue kindly re-check

prashanth-offcl:
@prashanthr6383 Currently in multiselect mode If I want to implement a behavior based on selection change then I have to listen to both itemClick (tree-view-item) and itemChange (tree-view). Ideally one event at the parent level should be enough and it should be emitted whenever there is a change wrt selection, be it keyboard or mouse click. 

After introducing itemChange the itemClick event is not really necessary but lets keep it for backward compatibility and have this documented in the storybook.

prashanth-offcl (approved):

PR #2634, opened 2024-06-24
NumberInput: Add decimal type and currency

### LINKED ISSUE #2440: Input: Add Currency Handling or new input type
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed feature requests
- [X] I have read the [Modus style guidelines](https://modus.trimble.com/components/web/introduction/)

### Proposal

Add ability for inputs to display/format/validate currency values. Implementation details TBD.

### Motivation and context

Need ability to present and format currency values.

### What is the issue regarding ?

@trimble-oss/modus-web-components

### Priority

Medium

### What product/project are you using Modus Components for ?

Modus

### What is your team/division name ?

Modus

### Are you willing to contribute ?

None

### Are you using Modus Web Components in production ?

_No response_

ElishaSamPeterPrabhu:
## Description

- With cleave-zen added decimal to help with localisation of numbers

References 
Fixes #2440 

## Type of change

- [x] Documentation update

## How Has This Been Tested?

https://deploy-preview-2634--moduswebcomponents.netlify.app/?path=/story/user-inputs-number-input--default

## Checklist


stencil-workspace/src/components/modus-number-input/modus-number-input.tsx
```diff
@@ -1,5 +1,5 @@
-// eslint-disable-next-line
```
coliff, line 1: don't remove this please

stencil-workspace/src/components/modus-number-input/modus-number-input.spec.ts
```diff
@@ -12,7 +12,7 @@ describe('modus-number-input', () => {
         <mock:shadow-root>
           <div class="modus-number-input">
             <div class="input-container medium" part="input-container">
-              <input class="text-align-left" id="mwc_id_0_number-input" tabindex="0" type="number">
```
coliff, line 15: If `type="text"` is being used for numeric input then should add `inputmode="numeric"` for virtual keyboards (Windows touchscreen, Chromebooks, iPads, smartphones etc).

stencil-workspace/src/components/modus-number-input/modus-number-input.tsx
ElishaSamPeterPrabhu: Added the comment back.

stencil-workspace/src/components/modus-number-input/modus-number-input.spec.ts
ElishaSamPeterPrabhu: Added `inputMode = "numeric"`

coliff: inputmode should be lowercase

coliff:
1. Please change inputmode to `decimal` (not `numeric`).
2. Please fix this accessibility error.
image

ElishaSamPeterPrabhu:
> 1. `decimal`

Including the cleave-zen library the input is required to have type text , hence the aria attributes like            
```
 aria-valuemax={this.maxValue}
 aria-valuemin={this.minValue}
 aria-valuenow={this.value} 
```
 are not valid , so can we remove them or is their any alternative? @coliff

coliff:
> > 1. `decimal`
> 
> Including the cleave-zen library the input is required to have type text , hence the aria attributes like
> 
> ```
>  aria-valuemax={this.maxValue}
>  aria-valuemin={this.minValue}
>  aria-valuenow={this.value} 
> ```
> 
> are not valid , so can we remove them or is their any alternative? @coliff

Yep, please remove them. They are not valid. Thanks!

ElishaSamPeterPrabhu:
> > Including the cleave-zen library the input is required to have type text , hence the aria attributes like
> > ```
> >  aria-valuemax={this.maxValue}
> >  aria-valuemin={this.minValue}
> >  aria-valuenow={this.value} 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > are not valid , so can we remove them or is their any alternative? @coliff
> 
> Yep, please remove them. They are not valid. Thanks!

Removed them.

coliff:
The number input PR is good in many ways, but I have some concerns...
Demo: https://deploy-preview-2634--moduswebcomponents.netlify.app/?path=/docs/user-inputs-number-input--default

- It uses the input type="text" now instead of type="number", so user can't use mouse scroll wheel to increase/decrease and the arrow buttons to adjust doesn't display. Maybe a minor thing?
- If you see Number Input Demo 1 - if this was for a credit card you wouldn't want commas or dots in the input. I think separators should be opt-in rather than on by default.
- DecimalPlaces is an optional attribute so I suggest it is off by default can be.
- Unable to enter a numeric input such as 0123456789 - this could be a valid use for a part number or something.

ElishaSamPeterPrabhu:
> The number input PR is good in many ways, but I have some concerns... Demo: https://deploy-preview-2634--moduswebcomponents.netlify.app/?path=/docs/user-inputs-number-input--default
> 
> * It uses the input type="text" now instead of type="number", so user can't use mouse scroll wheel to increase/decrease and the arrow buttons to adjust doesn't display. Maybe a minor thing?
> * If you see Number Input Demo 1 - if this was for a credit card you wouldn't want commas or dots in the input. I think separators should be opt-in rather than on by default.
> * DecimalPlaces is an optional attribute so I suggest it is off by default can be.
> * Unable to enter a numeric input such as 0123456789 - this could be a valid use for a part number or something.
> 
> I think to summarize, I think the defaults should be like what they were before and new functionality should be opt-in.... otherwise it could cause breaking changes / unintended side effects for users web apps.

Addressed all this comments

- Added manual scrolling ability along with plus or minus buttons to change number values.
- Removed commas or dots for the default version
- Set decimal places as optional attribute by removing default value as 2
- Enabled the ability to enter 0123456789, by adding property `is-credit-card`

coliff:
`is-credit-card` isn't the right name for that option. Credit cards can't begin with a zero anyway. I'm not sure what the best name would be... `allow-leading-zero` ?

ElishaSamPeterPrabhu:
> `is-credit-card` isn't the right name for that option. Credit cards can't begin with a zero anyway. I'm not sure what the best name would be though ... `allow-leading-zero` ?

That is a better name @cjwinsor, can we go with `allow-leading-zero` ?

cjwinsor:
I think we should be using the Intl.NumberFormat standard js object to handle formatting.  The end user would provide a locale and the number format options.  

I was looking at this library https://dm4t2.github.io/intl-number-input/guide/#installation but i wasn't keen on the format as you type aspect of it... but thats just my opinion.  If there was a way to use this library without the live formatting it would be great... otherwise it might be easier to just handle it ourselves.  I'm open to using another library if one is found as well that wraps the functionality in a useful way.  I just felt the picked library complicated the end user experience too much.

I'd like to reduce the complexity for the end user as we should expect that they will have or otherwise obtain the users locale.  

For the component, I envision something like 

```html
<modus-number-input type="currency" locale="en-US" currency="USD"></modus-number-input>
```

behind the scenes we are creating an input with type="text" (keep it as number otherwise), and we are doing something to the affect of
```js
Intl.NumberFormat(this.locale, { style: 'currency', currency: this.currency }).format(this.value);
```

We should probably also map any existing properties to the formatting that control the number of decimals.   Additional enhancements could be to restrict them from typing more decimals than they have the precision set to... I've seen libraries take a number with decimal places like 1234.5 and do something like 

```js
const formatter = new Intl.NumberFormat(this.locale);
const parts = formatter.formatToParts(1234.5);

let decimalPoint = '.';
let groupSeparator = ',';

for (const part of parts) {
  if (part.type === 'decimal') {
    decimalPoint = part.value;
  }
  if (part.type === 'group') {
    groupSeparator = part.value;
  }
}
```

Once you know those, you can do cool things like prevent them from typing more than max decimal digits after the decimalPoint.  


Additional thoughts are that we can leverage this for a `type=decimal`.   Additional props can be added as needed later on, but I think just `locale` and `currency` are enough to satisfy having a currency input initially.   Also helps we can use `navigator.language` to get the locale from the browser... not sure if modus consumers would want... but might be a good default if they don't pass in a locale.
@prashanth-offcl

prashanth-offcl (changes requested):
@ElishaSamPeterPrabhu Have you tried  Intl.NumberFormat or toLocalString method of Number? Also,  

- Looks like we are using the text input by default for all cases, better to keep the type as number if currency and locale props are not set
- Locale & currency symbol - The accepted values need to be documented properly. Make sure the values matches with the options in [toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString)  or create an internal mapping
- Step up & down behaviors
    - working in reverse when using scroll.
    - Not working with keyboard up and down arrow.
    - Not working without specifying locale.
    - The Prop `step` is not working.
- Currency prop is not working without locale prop.
- Min & Max value prop - Behavior changed, now it accepts any number and automatically sets the respective threshold if out of range
- Cleanup code - some places in e2e have references to Cleave package which is not used now.

prashanth-offcl:
Merged in #2864

Issue #2699, opened 2024-07-16
Utility Panel: Remove padding from slots


raymondboswel:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/trimble-oss/modus-web-components/blob/main/CONTRIBUTING.md)

### Describe the issue

Defining padding on the slot containers limits design freedom when using the Utility panel while not adding much value. It could be made configurable, but I think a BYOP (bring your own padding) would be best. 

### What is the issue regarding ?

@trimble-oss/modus-web-components

### What version of npm package are you using ?

0.34

### Priority

Medium

### What product/project are you using Modus Components for ?

Machine Playback / WorksManager

### What is your team/division name ?

Civil / Weka

### Are you willing to contribute ?

Yes

PR #2700, opened 2024-07-16
Utility panel: Remove hard coded padding

raymondboswel:
## Description

The current design does not provide flexibility to consumers who may want to use the full width of the panel. Removing the padding solves this.

References #2699

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

Storybook instance was started up, and styling inspected.

## Checklist


stencil-workspace/src/components/modus-utility-panel/modus-utility-panel.scss
```diff
@@ -34,13 +34,11 @@
       align-items: center;
       display: flex;
       height: 50px;
-      padding: 0 1rem;
```
prashanth-offcl, line 37: Can we keep the default behavior and provide an option to override the value using css variables?

raymondboswel: Doesn't that just complicate the API for little actual benefit? For a consumer to add padding to their own content is simple and natural. Having to look up a prop or set a css variable is considerably more effort. One of the main critiques for web components is difficulty in styling, thus being less opinionated is a step in the right direction IMO.

prashanth-offcl: @raymondboswel I made the suggestion with the following reasons in my mind sorry for not adding them. But I agree you are right and using CSS variables for style extensibility comes with a lot of drawbacks. 

- The default behavior with padding will benefit most of the integrators, only for edge cases we will have the need to remove the default padding. 
- If the value changes in future then only the web component needs to be updated.
- Removing padding is a breaking change for other integrators.

prashanth-offcl: @raymondboswel Based on the above comment I believe it is better to go with an optional property. could you please make that change?

PR #2723, opened 2024-07-24
Autocomplete: Handle dynamic options

### LINKED ISSUE #2710: Autocomplete: Handle dynamic options

ElishaSamPeterPrabhu:
## Description

- Added a prop `newOptions` to accept values dynamically

References 
Fixes #2710 

## Type of change

- [ ] Documentation update

## Checklist


cjwinsor:
My main feedback here is whether we can handle without adding a new prop, but instead allow `options` to also be the promise.  Not sure if this would require rethinking anything though.  I'd just want to say `autocomplete.options = getOptions;` where `getOptions` returns a promise when executed. 

@prashanth-offcl

prashanth-offcl:
> My main feedback here is whether we can handle without adding a new prop, but instead allow `options` to also be the promise. Not sure if this would require rethinking anything though. I'd just want to say `autocomplete.options = getOptions;` where `getOptions` returns a promise when executed.
> 
> @prashanth-offcl

Thanks @cjwinsor I'm thinking if we can introduce two new props loading and filterOptions that accepts an async function. 

If filterOptions prop is defined, then we will skip the filtering and call the method to get the filtered options from the integrator and display it. This will be useful for both dynamic options and custom filtering in static options if required. The integrator can use the loading prop to display the spinner while they fetch the data from server. The integrator will be responsible to reject promises in case of multiple calls to the filterOptions and ensure they resolve the most recent one.

prashanth-offcl (changes requested):
Waiting for Sam to try the discussed approach

prashanth-offcl (changes requested):
@ElishaSamPeterPrabhu some suggestions 

- The results in the list should be hidden when the loading prop is set to true
- Rename isLoading to loading, and getFilteredOptions to filterOptions
- The same behavior for customOptions is not handled.
- When changing the loading prop to true after resolving the promise, the no results found message is displayed before showing the filtered list. 
- The PR also seems to be contain code that needs to be cleaned up 
- Update documentation for filterOptions and loading prop

prashanth-offcl:
BTW I managed to test the updates in react test application using yalc and it works fine.

ElishaSamPeterPrabhu:
> @ElishaSamPeterPrabhu some suggestions
> 
> * The results in the list should be hidden when the loading prop is set to true
> * Rename isLoading to loading, and getFilteredOptions to filterOptions
> * The same behavior for customOptions is not handled.
> * When changing the loading prop to true after resolving the promise, the no results found message is displayed before showing the filtered list.
> * The PR also seems to be contain code that needs to be cleaned up
> * Update documentation for filterOptions and loading prop

All comments are addressed except custom options implementation and ''m not able to reproduce this
"When changing the loading prop to true after resolving the promise, the no results found message is displayed before showing the filtered list."

For custom options implementation are are we going to get the values and format separately or like in li format like
  `

    <li data-search-value="The Git Guru" data-id="1" style="padding: 8px">
      <div style="font-weight: bold">The Git Guru</div>
      <div style="font-size: 12px">Lead DevOps Engineer</div>
    </li>
    <li data-search-value="Bob the Builder" data-id="2" style="padding: 8px">
      <div style="font-weight: bold">Bob the Builder</div>
      <div style="font-size: 12px">Senior Construction Engineer</div>
    </li>
`

stencil-workspace/src/components/modus-autocomplete/modus-autocomplete.tsx
```diff
@@ -163,13 +183,20 @@ export class ModusAutocomplete {
...
 
   convertOptions(): void {
-    if (this.options && this.options.length > 0) {
-      if (typeof this.options[0] === 'string') {
-        this.options = this.options?.map((option) => ({
-          id: option,
-          value: option,
-        }));
-      }
+    if (Array.isArray(this.options) && this.options.length > 0) {
```
prashanth-offcl, line 186: can you revert this change? directly modifying the options without type check causes a loop.

prashanth-offcl (approved):

Issue #2197, opened 2024-02-23
Inactive Tabs Error


kuhnboy:
### Prerequisites

- [X] I have [searched](https://github.com/trimble-oss/modus-web-components/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/trimble-oss/modus-web-components/blob/main/CONTRIBUTING.md)

### Describe the issue

If no tab is set to active = true, click events are broken.

Example:
```html
<body>
    <modus-tabs></modus-tabs>
    <modus-tabs size="small"></modus-tabs>

    <script>
    const modusTabs = document.querySelector('modus-tabs');
    modusTabs.tabs = [
        {
        id: 5,
        label: 'Tab 1',
        },
        {
        id: 1,
        label: 'Tab 2',
        },
    ];
    modusTabs.addEventListener('tabChange', (evt) => {
        alert(JSON.stringify(evt.detail));
    });
    </script>
</body>
```

Upon click this results in:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'id')
    at o.handleTabChange (392.test.bundle.js:1:470)
```

### What operating system(s) are you seeing the problem on?

Windows

### What browser(s) are you seeing the problem on?

Microsoft Edge

### What is the issue regarding ?

@trimble-oss/modus-web-components

### What version of npm package are you using ?

0.16.0

### Priority

Medium

### What product/project are you using Modus Components for ?

ModusBlazor

### What is your team/division name ?

AECO

### Are you willing to contribute ?

None

PR #2872, opened 2024-09-18
Tab: Adding id as an "optional" params

roaldjap:
## Description

I added a functionality where `modusTabs.tabs` does not need to require to provide an `id` per tab instance - as label would be enough to generate a id for this to ensure `active` state are clicking even id were not provided.

Coincidentally, This also resolves an issue that relates to: https://github.com/trimble-oss/modus-web-components/issues/2197

References #
https://github.com/trimble-oss/modus-web-components/issues/2197

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

To reproduce this - initially I created a tab instance like this:
```
    const modusTabs = document.querySelector('modus-tabs');
    modusTabs.tabs = [
        {
          active: true,
          label: 'Tab 1',
        },
        {
           label: 'Tab 2',
        },
    ];
```

I noticed tabs were not changing its active state - even if `active: true` is set on first tab instance.

## Checklist

### DIFF
--- a/stencil-workspace/src/components/modus-tabs/modus-tabs.e2e.ts
+++ b/stencil-workspace/src/components/modus-tabs/modus-tabs.e2e.ts
@@ -57,11 +57,23 @@ describe('modus-tabs', () => {
     expect(tabChange).toHaveReceivedEvent();
   });
 
+  it('renders changes on tabs even without tab id provided', async () => {
+    const page = await newE2EPage();
+
+    await page.setContent('<modus-tabs></modus-tabs>');
+    const modusTabs = await page.find('modus-tabs');
+    modusTabs.setProperty('tabs', [{ label: 'Tab1' }, { id: 0, label: 'Tab 2' }]);
+    await page.waitForChanges();
+    const element = await page.find('modus-tabs >>> button');
+
+    expect(element.id).toEqual('tab-label-tab1');
+  });
+
   it('renders aria-label on tabs div when set', async () => {
     const page = await newE2EPage();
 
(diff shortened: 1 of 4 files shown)


prashanth-offcl (commented):
@roaldjap Can we use the index instead of generating a new ID? It’s challenging to identify the selected tab since the ID is created within the component when I want to show different tab content. Also it would be great if you can update the interface and storybook documentation.

vetrivel1 (commented):
stencil-workspace/src/components/modus-tabs/modus-tabs.tsx (line 88-93) The new code adds automatic ID generation for tabs when no ID is provided, but the kebabCase function doesn't handle null or undefined inputs safely. If tab.label is null or undefined, the function will throw an error when calling .match() on it. Consider adding a null check before calling kebabCase. 

stencil-workspace/src/utils/utils.ts (line 20-25) The kebabCase implementation assumes the input string will match the regex pattern and return an array, but it doesn't handle edge cases where the match returns null (empty strings, strings with only special characters). The .join() call would throw an error in such cases. 

stencil-workspace/src/components/modus-tabs/modus-tabs.e2e.ts (line 65) The test case uses a mixed approach for tab IDs (one tab without ID, one with numeric ID 0), but doesn't verify that both tabs render correctly. Consider expanding test coverage to validate both tabs. 

stencil-workspace/src/utils/utils.spec.ts (line 23-27) The unit test for kebabCase only tests a single case with whitespace. Add more test cases for different inputs like camelCase, PascalCase, and edge cases like empty strings or strings with special characters.

